PR #128, opened 2022-02-18
Bump minimum Python patch version

### LINKED ISSUE #127: Fix dependabot
Dependabot is complaining that the version of Python required by this package is incompatible with one of the dependencies, pandas:

```
  The current project's Python requirement (>=3.7,<4.0) is not compatible with some of the required packages Python requirement:
    - pandas requires Python >=3.7.1, so it will not be satisfied for Python >=3.7,<3.7.1
```

We should bump the version of Python accordingly. I don't know why pandas doesn't support 3.7.0, but I doubt many people are using the initial release anymore anyway.

Andy-Grigg:
This PR closes #127 

Dependabot is erroring because the advertised package Python compatibility is incompatible with pandas. We support 3.7.0 and up, whereas pandas supports 3.7.1 and up. The change here aligns these requirements.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ packages = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.7"
+python = "^3.7.1"
 
 # Packages for core library
 importlib_metadata = { version = ">=1.0", python  = "<3.8" }  # Granta MI STK requires 3.4.0


da1910:
I think this should be ok, do you know how far back we would need to go to support 3.7.0?

da1910:
> I think this should be ok, do you know how far back we would need to go to support 3.7.0?

It looks like there's some sort of bug with module imports in 3.7.0 - https://github.com/pandas-dev/pandas/issues/35309 
I think this should be fine

da1910 (approved):

PR #134, opened 2022-02-21
CI - Skip integration tests on dependabot PRs

da1910:
Skip integration tests when dependabot generates PRs


Andy-Grigg (approved):
Dependabot uses the `dependencies` label by default, which is what the CI/CD workflow expects to determine whether to skip integration tests or not. Deleting this row should mean dependabot creates the PR with the `dependencies` label instead.

PR #138, opened 2022-02-21
CI - Filter on branch name to skip dependabot integration tests

### LINKED ISSUE #137: The CI/CD workflow is not respecting the 'dependencies' label
When dependabot creates a PR, it adds the `dependencies` label to the PR. This should mean that integration tests and examples are not executed. However, this is currently not working as expected. All dependabot PRs run a full set of tests, even though the label is set.

This may be because the label is added after the PR is created. More investigation is necessary.

da1910:
Closes #137 

Use the branch name as the filter to determine when to skip integration tests on dependabot automated PRs

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -18,7 +18,7 @@ jobs:
   start_vm:
     name: Start Azure VM
     runs-on: ubuntu-latest
-    if: ${{ !contains( github.event.pull_request.labels.*.name, 'dependencies') }}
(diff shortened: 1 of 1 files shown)


Andy-Grigg (approved):
According to https://docs.github.com/en/actions/learn-github-actions/contexts#github-context this should work. The only way to test though is to merge this and re-run the dependabot checks.

da1910:
Now I think about it I should have prefixed this branch with dependabot!

Andy-Grigg:
Closes #137

PR #140, opened 2022-02-21
Fixup workflow trigger and skips

da1910:
The variable github.ref_name points to the pull request branch, not the source, so the previous change did not skip correctly. This uses the event information to get the source of the PR which does work.

Also simplify the logic around documentation building.


Andy-Grigg (approved):
These changes are as discussed. We'll now always run documentation tests for dependabot PRs even if the unit tests or style checks fail. This is fine, because these types of checks are very quick to run, and will give us a better picture of what the failures were.

Integration test execution is unaffected, as demonstrated by the checks on this PR.

Issue #136, opened 2022-02-21
Allow non-integration test runs of CI/CD workflow to run in parallel


Andy-Grigg:
The CI/CD workflow contains a `concurrency` section which enforces that one instance can only be active at once. This is required to ensure the integration test server is not started or stopped while another test is running.

However, this has the effect of also queueing up dependency PR checks that don't require access to the server. Github actions do allow specifying the `concurrency` on the `job` level, but this warning https://docs.github.com/en/actions/using-jobs/using-concurrency#overview concerns me; we could still end up shutting down the integration test VM before another workflow instance expects if we have multiple integration test instances.

A safer fix would be to have two separate workflows; one of which is used for unit tests only, and one that is used for integration tests. However, this would require a lot of duplication of workflow jobs.

PR #145, opened 2022-02-21
Split CI/CD workflow

Andy-Grigg:
Pre-requisite for #136, but will not solve it directly.

This PR separates the workflows into three distinct files:

1. `local_checks.yml`: Includes static checks, unit tests, wheel build, and docs build without examples.
2. `integration_checks.yml`: Starts the test server, runs integration tests and builds docs with examples. Then stops the test server
3. `release.yml`: Uploads the wheel to PyPI and creates a github release. All public-facing steps have been deleted for testing purposes.

- Workflow 1 is triggered based on PRs, pushes to main, and tags, as before. 
- Workflow 2 is triggered when Workflow 1 successfully completes on a non-dependabot branch, or when triggered manually. Workflow 2 has concurrency configured, so only a single instance can run at once
- Workflow 3 is triggered when Workflow 2 completes, but only executes if a tag has been pushed and the previous workflow successfully completed.

The recommended approach for doing this seems to be setting the dependent workflows to be triggered based on the completion of the prior workflow. I believe this means this PR needs to be merged so that Workflows 2 and 3 are in `main` and will trigger.

**Note: when this PR is merged, CI/CD will likely break, since it's unlikely that these changes will work first-time.**


da1910 (approved):
This looks superficially OK, let's see what happens

PR #153, opened 2022-02-22
Use Call-Workflow for CI/CD Concurrency Handling

Andy-Grigg:
This PR re-organizes the CI/CD workflows to ensure we have proper concurrency.

There are now two 'template' workflows:

- `local_checks.yml`: Runs style checks, unit tests, builds package, and builds docs without examples. No restrictions on concurrency.
- `server_checks.yml`: Starts the VM, runs integration tests, builds docs with examples, stops the VM. Can only run one instance at a time

The actual workflow then does the following:

- Run local checks
- If local checks pass and the run is not triggered by a PR to merge a branch that starts `dependabot/`, run server checks
- If server checks pass and we pushed a tag, release

The upshot of this is that we can have as many dependabot PRs in at once and the tests will run in parallel. For non-dependabot PRs, the local jobs will run in parallel, but the server checks will run sequentially.

See https://github.com/pyansys/grantami-bomanalytics/pull/154 for some examples of the dependabot versions of the workflow execution. 

Once this is merged and closed, we should also close https://github.com/pyansys/grantami-bomanalytics/pull/154. We should also update the required checks in the repo settings.


da1910 (approved):
This is quite clean, and it seems to work.

PR #159, opened 2022-02-23
Fix PyPI Documentation Links

### LINKED ISSUE #125: Shorten long line on README.md
The README.md contains a reasonably long line of example code. This looks ok on github, but results in a scrollbar on PyPI. We should reduce the length of this line to make the same code easier to read on PyPI.

### LINKED ISSUE #157: Fix/add documentation links
The documentation link in the README.md points to the general PyAnsys page, not to the specific page for this package

Also, we should add a documentation link to the pyproject.toml so the link appears on the PyPI page.

Andy-Grigg:
Closes #157 

This PR adds some more context to the documentation link in the README, which ends up on PyPI. It also adds a documentation tag to the pyproject.toml, to ensure the documentation link is added to the sidebar on PyPI.

Closes #125

Also shorten the long line in the quick code section by removing the keyword.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -31,7 +31,8 @@ Alternatively, clone and install with:
 
 Documentation
 -------------
-`PyAnsys <https://grantami.docs.pyansys.com>`_
+See `Granta MI BoM Analytics Documentation <https://grantami.docs.pyansys.com>`_
+for more details.
 
 
 Usage
@@ -44,7 +45,7 @@ Here's a brief example of how the package works:
 
     >>> from pprint import pprint
     >>> from ansys.grantami.bomanalytics import Connection, queries
-    >>> cxn = Connection(servicelayer_url='http://localhost/mi_servicelayer').with_autologon().connect()
+    >>> cxn = Connection("http://my_grantami_server/mi_servicelayer").with_autologon().connect()
     >>> query = (
     ...     queries.MaterialImpactedSubstancesQuery()
     ...     .with_material_ids(['plastic-abs-pvc-flame'])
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -31,7 +31,8 @@ Alternatively, clone and install with:
 
 Documentation
 -------------
-`PyAnsys <https://grantami.docs.pyansys.com>`_
```
da1910, line 34: This seems like a long hyperlink, but it matches the style and I can't see a more sensible place to start the link.

README.rst
```diff
@@ -44,7 +45,7 @@ Here's a brief example of how the package works:
 
     >>> from pprint import pprint
     >>> from ansys.grantami.bomanalytics import Connection, queries
-    >>> cxn = Connection(servicelayer_url='http://localhost/mi_servicelayer').with_autologon().connect()
```
da1910, line 47: I think this should, if possible, match the documentation. There we leave the keyword but add a line break before `.with_autologon()`. Either it should be that, or we should change the documentation to match.

Andy-Grigg: My motivation in doing it this way was to show the simplest possible code.

Having the line break syntax makes it look more complicated than it actually is to create a connection if someone's running in Windows and wants to authenticate with Kerberos, which is going to be the case for probably 80%+ of the people using this package.

It's a bit inconsistent, yes, but I think the goals of this code is different. Here we're showing the simplest possible working example to give a _potential_ user of the package a general idea of what the interface looks like. The examples though are to educate an _actual_ user as to how they use the package.

da1910: OK, I think that makes sense

da1910 (approved):

PR #161, opened 2022-02-24
[CI]156 - Add generate_release_notes flag to release

### LINKED ISSUE #156: Use generate_release_notes
Consider using `generate_release_notes` as described in:
https://github.com/softprops/action-gh-release#-customizing

Apply it here:

https://github.com/pyansys/grantami-bomanalytics/blob/087c596b31291cb4106edb2d20823b72e87f581c/.github/workflows/ci_cd.yml#L91-L97

Will be especially helpful for patch/minor releases.

da1910:
Closes #156 

Add the `generate_release_flag` argument to the release action. This will automatically summarise contributions to a release on the github releases page.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -91,6 +91,7 @@ jobs:
       - name: Release
         uses: softprops/action-gh-release@v1
         with:
+          generate_release_notes: true
           files: |
             ~/**/*.whl
             ./**/*.zip


Andy-Grigg (approved):
Have we thought about whether we should customize the release template? https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#configuration-options

Maybe it's best to autogenerate with a default template first and see what it looks like. We can then customize for subsequent releases.

PR #213, opened 2022-07-19
Add dummy examples

### LINKED ISSUE #164: Create a 'dummy' example
Currently when dependabot creates PRs, the CI/CD that runs skips examples. When the examples are skipped, the `conf.py` creates a placeholder `index.rst` and includes it in the source folder. However, this has the effect of not only skipping requiring a connection to the server, but also means we don't test example creation at all.

The execution of examples here is non-trivial, we use an IPython environment with some monkeypatches to force a connection to the integration test server, regardless of what is specified in the example. We should create a dummy example that checks the correct behavior of this IPython environment. That way, when examples are skipped, there will still be some evidence as to whether they would have built correctly.

Andy-Grigg:
Closes #164 

Creates two example files that test the display-related functionality in the examples. These files go through the same process as the real examples but do not require a live Granta MI instance, so they run much quicker than the real examples.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -204,8 +204,11 @@ epub_exclude_files = ["search.html"]
 cwd = Path(os.getcwd())
 assert (cwd.name == "source")
 EXAMPLES_DIR_NAME = "examples"
+DUMMY_EXAMPLES_DIR_NAME = "examples-dummy"
+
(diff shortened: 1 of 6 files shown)


Andy-Grigg:
There seems to be a problem with the test machine returning 502 responses. The dummy examples work, but I'd like to confirm the real examples still work before merging.

ludovicsteinbach (approved):
Looks good

PR #231, opened 2022-10-17
DOCS - Use ansys-sphinx-theme

da1910:
Use ansys-sphinx-theme in place of deprecated pyansys-sphinx-theme.

This is part of a pyansys wide initiative to use a single documentation theme.


Andy-Grigg:
This may require some work upstream on the theme before we can adopt it, the dataframe working is wonky in some places:

image

Andy-Grigg:
Fixed locally with a CSS override. The change seems to have been to set `table-width: 100%` for dataframes, which is good and allows the element to fill all available space.

However, nbsphinx adds a whole load of in-html CSS after the linked css, and adds a `table-layout: fixed`, which forces fixed equal width columns. As a result, the fix is to add a CSS file with an `!important` statement. Makes me sad, but there's no other option.

Just waiting for the build to complete to verify that it's really fixed.

Andy-Grigg:
image

Resolved with latest commits. I'll raise a PR to see about getting this merged into the theme, but this fix will work right now, and shouldn't impact things if the fix does get merged upstream.

doc/source/_static/css/dataframe.css
```diff
@@ -0,0 +1,10 @@
+table.dataframe {
+    table-layout: auto !important;
+}
+
+/*
+!important is unavoidable
```
da1910, line 6: It is sometimes unavoidable...

da1910:
Changes all look as I'd expect 👍

Andy-Grigg (approved):
Approved by @da1910 in comment above.

PR #239, opened 2022-11-17
Add client connection check

### LINKED ISSUE #215: Handle missing BoM Analytics service better
### 📝 Description of the feature

If a user tries to use this package against an instance of MI without BoM Services installed, they get a 404 error. it would be better if we checked for the existence of the package during the authentication step, and gave a more helpful error message if it is missing.

### 💡 Steps for implementing the feature

_No response_

### 🔗 Useful links and references

_No response_

Andy-Grigg:
Closes #215 

Adds a check as part of the `connect` method to perform a test query. The BoM Analytics service provides the Yaml endpoint which just returns the docs, so the test checks that this query can be performed without raising an exception.

If an exception is raised, then separate error messages are raised depending on whether it's a 404 (service not installed) or something else (probably a 500, check the service layer logs).

### DIFF
--- a/src/ansys/grantami/bomanalytics/_connection.py
+++ b/src/ansys/grantami/bomanalytics/_connection.py
@@ -323,11 +323,11 @@ class BomAnalyticsClient(ApiClient):
         ...
 
     @overload
-    def run(self, query: Yaml) -> str:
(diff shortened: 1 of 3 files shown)


ludovicsteinbach (approved):
Looks good to me

PR #241, opened 2022-11-21
238 - Migrate to dependency groups

### LINKED ISSUE #75: Re-instate development installation instructions following poetry update
Related to #72 

Poetry has support for development (`pip install -e .`) installs in, but a new release has not been created yet which includes this addition.

https://github.com/python-poetry/poetry/issues/34

Once this feature has been released, we should revert the documentation to cover development installs again.

### LINKED ISSUE #238: Re-organize test and docs extras into poetry groups

da1910:
Closes #238 
Closes #75 

Reorganise dev dependencies into a main `dev` group, and a `doc` group. Removes defined extras, since none are runtime dependencies. Consolidates `test` and `dev` group into one group, since at present tox-poetry does not support the `--with` option, and we never actually installed one without the other.


pyproject.toml
```diff
@@ -43,82 +43,54 @@ importlib_metadata = { version = ">=1.0", python  = "<3.8" }  # Granta MI STK re
...
-  "sphinx-toolbox",
-  "jupyterlab",
-  "jupytext",
-  "lxml",
-  "tabulate",
-  "pandas",
-  "nbsphinx",
-]
-
-examples = [
```
Andy-Grigg, line 98: I think we should leave this extra in - it's useful for end users who want to install all the requirements for running the jupyter notebooks.

Andy-Grigg (changes requested):
All looks good, one change needed to keep the examples extra around.

Andy-Grigg (approved):
wheel can be installed with and without [examples] - adding the extra adds jupyter and all the various requirements

Issue #212, opened 2022-07-15
Add 'Getting started' page to the documentation


Andy-Grigg:
### 📝 Description of the feature

Add a 'Getting started' section to the docs, covering:

- Installing the package
- Enabling OIDC authentication
- Ansys Granta software requirements
- Verifying the installation

Use the pyMAPDL documentation as a starting off point

PR #243, opened 2022-11-22
#212 - Getting Started Section

da1910:
This PR uses the format of the PyMAPDL Getting Started section to add more important information to the documentation. Covers platform support, python versions, installation instructions and a means to verify the installation.

### DIFF
--- a/doc/source/index/getting_started.rst
+++ b/doc/source/index/getting_started.rst
@@ -1,0 +1,55 @@
+Getting Started
+---------------
+To use this package you will require access to a Granta MI server
+that includes MI Restricted Substances Reports 2022 R1
+or later.
+
+Installation
+~~~~~~~~~~~~
+The ``ansys.grantami.bomanalytics`` package currently supports Python 3.7 through 3.10 on Windows and Linux.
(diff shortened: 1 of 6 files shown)


doc/source/index/getting_started.rst
```diff
@@ -0,0 +1,55 @@
...
+
+    pip install git:https://github.com/pyansys/grantami-bomanalytics.git
+
+For a local "development" version, install with git and poetry:
+
+.. code::
+
+    git clone https://github.com/pyansys/grantami-bomanalytics
+    cd grantami-bomanalytics
+    poetry install
```
Andy-Grigg, line 29: I believe as of poetry 1.2 `pip install -e .` works, but we should probably be advising people to use poetry. I think this is fine.

Andy-Grigg (approved):
LGTM

doc/source/index/getting_started.rst
da1910: If you install using pip, as far as I know you won't get any dependency groups, you'll only get the main dependencies.

PR #245, opened 2022-11-24
Docs - Remove sphinx_autodoc_typehints dependency

ludovicsteinbach:
This PR removes the dependency on sphinx_autodoc_typehints. It was only used for removing typehints from signatures in the docs, which is now achieved by sphinx.ext.autodoc
Uses `autodoc_type_aliases` to prevent the fully qualified name of `Yaml` to be output.


Andy-Grigg (approved):
Had a look through the API documentation, everything looks good

PR #248, opened 2022-12-13
Pin tox to 3.x in local checks

da1910:
Tox v4 is out and is incompatible with some legacy CLI options, it also doesn't work with tox-gh-actions yet. Pin to v3


Andy-Grigg (approved):
👍

PR #262, opened 2023-02-10
Reformat for new black version

Andy-Grigg:
Black 23.x includes changes to the formatting rules, which is breaking CI. This change fixes a formatting error raised by Black 23.


Andy-Grigg:
Azure credentials have expired, so the test VM cannot be started. I have requested new credentials.

da1910 (approved):
Looks familiar!

PR #274, opened 2023-03-13
Enable multi version docs

ludovicsteinbach:
Enables multi version docs:
- Cherry picking the two commits from PRs on main, after manually upgrading the theme version to minimize poetry.lock merge conflicts
- Bump the package version


Andy-Grigg (approved):
All looks as expected. Using the pyansys actions cuts down on the amount of custom yaml, which is definitely a good thing!

PR #275, opened 2023-03-14
Bump version to 1.2.0

ludovicsteinbach:
Should help clarify the new docs, which are now a bit confusing since dev and stable are on the same minor version


Andy-Grigg (approved):
👍

PR #276, opened 2023-03-14
Fix GitHub release

ludovicsteinbach:
- Fix docs url in pyproject
- Fix condition for dev docs deploy
- Fix gh release step: see https://github.com/pyansys/grantami-bomanalytics/releases/tag/untagged-bb564b6493b96d77faba, generated in CI before putting some conditions back

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -53,7 +53,7 @@ jobs:
 
   Release:
     if: contains(github.ref, 'refs/tags') && github.event_name == 'push'
-    needs: doc-deploy-dev
+    needs: integration_checks
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -43,7 +43,7 @@ jobs:
     name: "Deploy development documentation"
     runs-on: ubuntu-latest
     needs: [ integration_checks ]
-    if: github.event_name == 'push'
+    if: github.event_name == 'push' && !contains(github.ref, 'refs/tags')
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_cd.yml
```diff
@@ -71,28 +71,43 @@ jobs:
...
-        uses: pyansys/actions/doc-deploy-stable@v4
+      - uses: actions/download-artifact@v3
         with:
-          cname: ${{ env.DOCUMENTATION_CNAME }}
-          token: ${{ secrets.GITHUB_TOKEN }}
-          doc-artifact-name: Documentation-html
+          name: Documentation-html
+          path: ~/dist
+
+      - name: "Compressing HTML documentation"
```
da1910, line 81: You shouldn't need to explicitely zip artifacts, if you mark a folder for upload then it should be zipped by default

ludovicsteinbach: Wouldn't downloading it unzip it then?

.github/workflows/ci_cd.yml
```diff
@@ -91,8 +103,11 @@ jobs:
       - name: Release
         uses: softprops/action-gh-release@v1
         with:
+          prerelease: true
```
Andy-Grigg: Presumably these three additions will be removed before merge.

Andy-Grigg (approved):
All seems good, and the draft release looks right. I don't see anything in the release docs https://github.com/softprops/action-gh-release that explicitly states that it will zip a folder, but it would be nicer if we didn't need to rely on any more actions than absolutely necessary.

.github/workflows/ci_cd.yml
ludovicsteinbach: Indeed

PR #282, opened 2023-03-15
Set ActAsReadUser header

### LINKED ISSUE #277: Error messages are returned for duplicate substance records, even though one is withdrawn
### 🔍 Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### 🐞 Description of the bug

When running an ImpactedSubstances analysis, sometimes I see messages indicating multiple substance records have been matched by a single row. However, I only see one record in MI Viewer.

Under closer inspection, I realize that there is an additional record, but it is withdrawn, as a result of following the duplicate CAS proceedure.

### 📝 Steps to reproduce

1. Create a new Restricted Substance record with the same CAS number as an existing record which is included in a material tabular attribute
2. Ensure the new RS record is impacted by the same legislations as the original record
3. Withdraw the new RS record
4. Run an ImpactedSubstance query on the material attribute, including one of the legislations impacting both substance records
5. Observe the message warning about duplicate records

### 💻 Which operating system are you using?

Windows

### 📀 Which ANSYS version are you using?

2023R1

### 🐍 Which Python version are you using?

3.10

### 📦 Installed packages

```shell
ansys-grantami-bomanalytics @ file:///C:/path/grantami-bomanalytics
ansys-grantami-bomanalytics-openapi==1.0.0
ansys-openapi-common==1.2.1
certifi==2022.12.7
cffi==1.15.1
charset-normalizer==3.1.0
cryptography==39.0.2
idna==3.4
pycparser==2.21
pyparsing==3.0.9
pypiwin32==223
pyspnego==0.8.0
python-dateutil==2.8.2
pywin32==305
requests==2.28.2
requests-negotiate-sspi==0.5.2
requests-ntlm==1.2.0
six==1.16.0
urllib3==1.26.15
```

ludovicsteinbach:
Closes #277 

I could reproduce the issue following the detailed steps provided. Assessment in the issue is correct, setting the header fixes the issue.
~~I have not added integration tests, as I could not find existing records that would result in the issue, but confirmed fix with a local database.~~
So far decided against making this configurable, since we can't see any use cases for it.

Added an integration test, which required added a new user that would have been able to see the withdrawn record.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -38,6 +38,8 @@ jobs:
       TEST_SERVER_URL: ${{ secrets.TEST_SERVER_URL }}
       TEST_SERVER_READ_USER: ${{ secrets.TEST_SERVER_READ_USER }}
       TEST_SERVER_READ_PASS: ${{ secrets.TEST_SERVER_READ_PASS }}
+      TEST_SERVER_WRITE_USER: ${{ secrets.TEST_SERVER_WRITE_USER }}
+      TEST_SERVER_WRITE_PASS: ${{ secrets.TEST_SERVER_WRITE_PASS }}
 
   doc-deploy-dev:
     name: "Deploy development documentation"
--- a/.github/workflows/server_checks.yml
+++ b/.github/workflows/server_checks.yml
@@ -17,6 +17,10 @@ on:
         required: true
       TEST_SERVER_READ_PASS:
         required: true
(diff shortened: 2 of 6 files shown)


Andy-Grigg (approved):
Change looks as expected. I'd like to have an integration test if possible, can you add and withdraw an appropriate record on the test db on the server?

da1910 (approved):
Looks as I expected, and the tests look to cover the functionality.

PR #285, opened 2023-03-16
Fix pypi upload wheel path

ludovicsteinbach:
1.1.3 release failed because the upload to pypi is misconfigured after the changes made in #276


da1910 (approved):
Path looks good

PR #286, opened 2023-03-16
Re-order release steps

ludovicsteinbach:
Re-ordered release steps. Action to deploy the docs was deleting the downloaded artifacts, preventing twine from finding them


da1910:
How helpful of that action... Perhaps we can raise an issue to get that fixed! Otherwise looks as intended.

PR #312, opened 2023-05-24
DOC - Update intersphinx mapping to 3.11 not dev

da1910:
The dev branch of python has renamed the builtin boolean reference, this breaks the docs build. Pin to 3.11 which works fine at present.


Andy-Grigg (approved):
👍

PR #337, opened 2023-08-17
CI - Minor script fixes

da1910:
Couple of minor fixes:
1. Fix name of database in modify_custom_rs_db.py - we don't fetch tables after renaming, so we need the old name
2. Only use history GUID to avoid issues with new versions


ludovicsteinbach (approved):
LGTM

PR #343, opened 2023-08-24
#291 Fix EU REACH, and some other miscellaneous documentation improvements

### LINKED ISSUE #291: Update Basic usage example
### Description of the modifications

The `Basic usage example` refers to "REACH - The Candidate List", which apparently does not exist in RS 2023 R1 anymore. Users running the example would get no impacted substances in the query results

### Useful links and references

https://bomanalytics.grantami.docs.pyansys.com/version/stable/examples/0_Getting_started.html?highlight=reach

Andy-Grigg:
Closes #291 

After the move to use the latest version of the Restricted Substances database for CI, we now see the effects of changing from 'REACH' to 'EU REACH':

https://bomanalytics.grantami.docs.pyansys.com/version/dev/examples/0_Getting_started.html#View-query-results

image

The REACH legislation 'Short title' was changed from 'REACH - The Candidate List' to 'EU REACH - The Candidate List' to disambiguate from UK REACH, China REACH, Korea REACH, etc. As a result, the examples now return zero results.

This PR replaces 'REACH' with 'EU REACH' everywhere, which should reinstate the results. Check the built docs in the pre-merge checks to confirm.

I have also made a few other miscellaneous documentation changes while I was in the code:

* Fixed a missing space in a connection error string
* Replaced 'localhost' with 'my_server_name', to be consistent with jupyter notebook examples
* Added clarification as to what should be used for the legislation_name parameter. Note that we might change this to use the legislation id instead. At this point we could change the parameter name, causing people to adjust their scripts to avoid a silent failure.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -49,7 +49,7 @@ Here's a brief example of how to use PyGranta BoM Analytics:
     >>> query = (
     ...     queries.MaterialImpactedSubstancesQuery()
     ...     .with_material_ids(['plastic-abs-pvc-flame'])
-    ...     .with_legislations(['REACH - The Candidate List'])
+    ...     .with_legislations(['EU REACH - The Candidate List'])
     ... )
 
     # Print out the result from the query.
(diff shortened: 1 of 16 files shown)


Andy-Grigg:
The built examples contain a populated list of impacted substances:

image

da1910 (approved):

PR #345, opened 2023-08-25
MAINT - Bump dependencies

da1910:
Run poetry update on the project, relax pins to carets and bump ipython to 8.x

### DIFF
--- a/examples/0_Getting_started.py
+++ b/examples/0_Getting_started.py
@@ -93,7 +93,11 @@ query
 # consolidate the cells above into a single step:
 
 # + tags=[]
-query = queries.MaterialImpactedSubstancesQuery().with_material_ids(["plastic-abs-high-impact"]).with_legislations(["EU REACH - The Candidate List"])  # noqa: E501
+query = (
+    queries.MaterialImpactedSubstancesQuery()
+    .with_material_ids(["plastic-abs-high-impact"])
+    .with_legislations(["EU REACH - The Candidate List"])
+)  # noqa: E501
 query
 # -
 
--- a/examples/0_Getting_started.py
+++ b/examples/0_Getting_started.py
@@ -93,11 +93,7 @@ query
 # consolidate the cells above into a single step:
 
 # + tags=[]
-query = (
-    queries.MaterialImpactedSubstancesQuery()
-    .with_material_ids(["plastic-abs-high-impact"])
-    .with_legislations(["EU REACH - The Candidate List"])
-)  # noqa: E501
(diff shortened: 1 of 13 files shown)


examples/0_Getting_started.py
```diff
@@ -93,7 +93,11 @@
 # consolidate the cells above into a single step:
 
 # + tags=[]
-query = queries.MaterialImpactedSubstancesQuery().with_material_ids(["plastic-abs-high-impact"]).with_legislations(["EU REACH - The Candidate List"])  # noqa: E501
```
Andy-Grigg, line 96: I deliberately haven't run black on this cell in the past because the syntax is a bit weird for non-experienced Python users, at least until it's introduced in the cell following. Could you revert and add `# fmt: off` and `# fmt: on` around this cell?

examples/1_Impacted_Substances_Queries/1-1_Materials_impacted_substances.py
```diff
@@ -52,9 +52,7 @@
 from ansys.grantami.bomanalytics import queries
 
 mat_query = (
-    queries.MaterialImpactedSubstancesQuery()
```
Andy-Grigg, line 55: Now I think about it, I think I formatted these other files to be narrower than what black wanted because otherwise there's a lot of horizontal scrolling in the HTML docs.

Maybe we just need to exclude the notebook python scripts from black altogether, and manage it manually.

pyproject.toml
```diff
@@ -43,36 +43,36 @@ ansys-openapi-common = "< 2.0.0"
 ansys-grantami-bomanalytics-openapi = "1.1.0"
 
 # Packages for the examples extra
-jupyterlab = { version = "3.6.5", optional = true }
-lxml = { version = "4.9.3", optional = true }
-pandas = { version = "1.3.5", optional = true }
-tabulate = { version = "0.9.0", optional = true }
+jupyterlab = { version = "^3.6.5", optional = true }
+lxml = { version = "^4.9.3", optional = true }
+pandas = { version = "^1.5.0", optional = true }
```
Andy-Grigg, line 48: Can we use pandas 2.0.0? Or is this a Python 3.8 thing?

Andy-Grigg (changes requested):

pyproject.toml
da1910: I think we can, let me see

examples/1_Impacted_Substances_Queries/1-1_Materials_impacted_substances.py
da1910: They _should_ be excluded, I think something weird happened with black when I updated pre-commit. I've reverted said changes, let's see if they get "fixed" again

Andy-Grigg (approved):

PR #348, opened 2023-08-25
347 - Use regex to replace server URL

### LINKED ISSUE #347: The repr in the 'Database-specific configuration' example is incorrect
### Description of the modifications

The repr should contain the spec-to-spec linking configuration if it is set. This is as a result of us mocking the repr to hide the actual server name.

### Useful links and references

https://bomanalytics.grantami.docs.pyansys.com/version/dev/examples/3_Advanced_Topics/3-3_Database-specific_configuration.html#Specification-to-Specification-links

Andy-Grigg:
Closes #347 

Use a regular expression to replace the URL in the BomAnalyticsClient repr. Raise a ValueError if no match is found.

### DIFF
--- a/.ipython/profile_default/startup/10-monkeypatch-connection.py
+++ b/.ipython/profile_default/startup/10-monkeypatch-connection.py
@@ -35,13 +35,15 @@ Connection.with_autologon = with_autologon
 server_url = "http://my_grantami_server/mi_servicelayer"
 
 BomAnalyticsClient._original_repr = BomAnalyticsClient.__repr__
-regex = re.compile(r'url="(.)+"')
+# \S is character class for 'not whitespace'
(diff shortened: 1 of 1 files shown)


.ipython/profile_default/startup/10-monkeypatch-connection.py
```diff
@@ -30,12 +31,20 @@ def with_autologon(self: Connection) -> Connection:
 Connection.with_autologon = with_autologon
 
 
-# Monkey patch the Client object to report the url specified in the code, or the one below if not overridden
+# Monkey patch the Client object to report the url specified below
```
Andy-Grigg, line 34: I can't see where this would be overridden, so I changed the comment

da1910 (approved):
Looks good, I didn't know about re.subn, you learn something every day!

PR #356, opened 2023-08-31
CI - Report Coverage to Codacy

da1910:
Codecov seems to be out of favor right now, codacy seems to be the place to be, so lets report there if we can.

### DIFF
--- a/.github/workflows/local_checks.yml
+++ b/.github/workflows/local_checks.yml
@@ -60,6 +60,12 @@ jobs:
       - name: Test with tox (no integration tests)
         run: tox -- -m "not integration"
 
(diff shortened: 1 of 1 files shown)


Andy-Grigg:
@da1910 Is this still relevant?

da1910:
I don't actually know, we should check with the Pyansys team what the current recommendation is. I don't actually mind which one we pick, but we should be reporting somewhere!

Andy-Grigg:
Coverage seems to be used most generally, but agreed, it would be good to have an agreed-upon direction.

Andy-Grigg:
We seem to have converged on codecov. Closing this PR.

PR #386, opened 2024-01-11
Merge from grantami-bomanalytics-private

Andy-Grigg:
Incorporate all development from private fork

Obviously there's a huge number of commits here, that represent the complete development of the sustainability portion of this package. I have gone through the workflow files and cleaned out any private-specific additions though, so hopefully CI will complete.


Andy-Grigg:
Updated all relevant secrets, CI seems to be passing.

ludovicsteinbach:
It looks reasonable from a quick spot check on some files.
Codacy raises valid points that we should eventually address. CI is failing when cloning the repo, which would go away if we switched back to using a dependency from PyPI

Andy-Grigg:
Updated to use the v2.0.0 of bomanalytics-openapi

See #388 for some codacy fixes. The others are either not real errors or more difficult to fix at this point

ludovicsteinbach (approved):

ludovicsteinbach:
I think we'll have to put back the steps to allow long paths in CI. Filenames in the openapi package are just too long, even when the dependency is sourced from PyPI

Andy-Grigg:
Merging as-is, the codacy issues are acceptable.

Issue #389, opened 2024-01-12
Re-use the README.rst file in the main documentation


Andy-Grigg:
The RecordLists repo reuses elements of the README.rst file in the documentation, see https://raw.githubusercontent.com/ansys/grantami-recordlists/main/doc/source/index.rst for an example

We should consider if we can do the same thing here

PR #390, opened 2024-01-12
Add metapackage instructions

Andy-Grigg:
Add instructions for installing this package via the pygranta metapackage.

There isn't as much reuse in this repo as in recordlists, and so the change was required in two places. I have created #389 to see if we can address this in future.


ludovicsteinbach (approved):
Looks good

PR #396, opened 2024-01-16
Address vale issues

Andy-Grigg:
Fix two vale issues


Andy-Grigg:
Cancelled CI until everything moves through, but the doc style check went through.

ludovicsteinbach (approved):
Looks ok to me.

PR #413, opened 2024-01-19
Add labeler workflow

ludovicsteinbach:
Add a workflow to automatically label PRs. There will be a follow up PR to use the labels to organize automatically generated release notes.

The configuration is identical to recordlists, which is the default that comes with ansys repo templates.
Labels are applied based on two things:
- location of the modified files
- name of the branch (`maint/*` -> `maintenance`, etc)

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,6 +1,6 @@
 documentation:
 - changed-files:
-  - any-glob-to-any-file: ['doc/source/**/*']
+  - any-glob-to-any-file: ['doc/source/**/*', 'examples/**/*']
 maintenance:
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,9 @@
+documentation:
+- changed-files:
+  - any-glob-to-any-file: ['doc/source/**/*']
```
Andy-Grigg, line 3: Should probably add examples to this as well?

Andy-Grigg (approved):
Looks fine, but I think we'll only know once it's merged and running

PR #443, opened 2024-03-12
Update to support 2024 R2

### LINKED ISSUE #427: Test PyGranta BoMAnalytics against 2024 R2
Once we get closer to release we should test against 2024 R2. This can be achieved by pointing the `main` branch at the new 'next' test VM.

Andy-Grigg:
Closes #427 

No changes to the underlying BoM Analytics services that should impact us, but openapi-common has gone to v2, which we will need to support to ensure compatibility with serverapi-openapi

Steps to get this PR passing:

* [x] Install 2024 R2 reports on server
* [x] Update package to include openapi-common v2 and bomanalytics-openapi alpha
* [x] Create 2024 R2 RS database and add to Next test server
* [x] Type hint fixes - this includes both changes to get mypy passing with type-hinted dependencies, and the changes introduced in openapi-common v2 to support Unset


tests/inputs/examples.py
```diff
@@ -1,72 +1,89 @@
...
 }
 
-examples_as_strings = {k: html.unescape(v) for k, v in _example_dict_raw.items() if v}
-examples_as_dicts = {k: json.loads(v) for k, v in examples_as_strings.items() if v}
+ID = "<ID>"
+BOM_PART_RECORD_HISTORY_IDENTITY = "92758"
+
+
+def fill_gaps(dct: Dict[str, Any]):
```
Andy-Grigg, line 75: This is all ultimately due to some examples being incomplete, which now fail to deserialize following the openapi-common v2 update because they are not marked as nullable in the spec.

This hook fills them in with arbitrary values, such that they can be correctly deserialized and validated

tests/test_integration_tests.py
```diff
@@ -14,7 +14,10 @@
 
 indicators = list(INDICATORS.values())
 
+out_of_date_database_xfail = pytest.mark.xfail(reason="Out of date database", raises=ValueError, strict=True)
```
Andy-Grigg, line 17: Tests currently fail because the database is out of date.

This should be a `GrantaMIException`. It's surfacing as a `ValueError` because the deserialization is failing, again as a result of the v2 openapi-common upgrade. When a critical-error occurs, fields are null which shouldn't be according to the spec, and so deserialization fails. See CR-1489 for details.

Andy-Grigg: I've updated the database, tests are still failing because of other non-nullable fields which are coming back from the server as null.

Andy-Grigg: This is fixed

src/ansys/grantami/bomanalytics/_item_results.py
```diff
@@ -56,7 +58,7 @@ def create_material_impacted_substances_result(
         item_result = MaterialWithImpactedSubstancesResult(
             reference_type=reference_type,
             reference_value=result_with_impacted_substances.reference_value,
-            legislations=result_with_impacted_substances.legislations,
+            legislations=_cast_unset_union_to_any(result_with_impacted_substances.legislations),
```
Andy-Grigg, line 61: Lots of changes here, mainly related to either converting things that are optional to things that we want to be nullable in the idiomatic layer, or just stripping the optional 'Unset_Type' type from things that we know are actually required, but just not specified as such in the API

Andy-Grigg: My chosen approach was to do this as high up in the call stack as possible, to try and avoid changing signatures of functions that we know are good.

src/ansys/grantami/bomanalytics/queries.py
```diff
@@ -514,7 +535,11 @@ def _call_api(self, api_method: Callable[..., models.ModelBase], arguments: Dict
     @abstractmethod
     def _run_query(
         self,
-        api_instance: Union[api.ComplianceApi, api.ImpactedSubstancesApi, api.SustainabilityApi],
+        api_instance: Union[  # type: ignore[override]
```
Andy-Grigg, line 538: I couldn't figure out how to make a type variable for this method with the right contravariant property, so I just ignored the warning

tests/test_integration_tests.py
```diff
@@ -274,9 +275,9 @@ def test_sustainability_summary_query(self, connection):
         assert beryllium_summary.mass_before_processing.value == pytest.approx(0.027)
         assert beryllium_summary.material_reference.record_guid is not None
         assert beryllium_summary.climate_change.value == pytest.approx(15.52, DEFAULT_TOLERANCE)
-        assert beryllium_summary.climate_change_percentage == pytest.approx(54.32, DEFAULT_TOLERANCE)
+        assert beryllium_summary.climate_change_percentage == pytest.approx(49.25, DEFAULT_TOLERANCE)
```
Andy-Grigg, line 278: A lot of these values changed, I assume because of changes in the database. I will double check though, don't merge until this thread is resolved.

tests/inputs/examples.py
Andy-Grigg: The null values have been removed in a subsequent PR - we should re-instate model and API examples in the openapi package and copy those in here

tests/test_integration_tests.py
Andy-Grigg: I haven't gone through all the calculations manually, but I have checked the databases and there are differences in eco indicators between versions, so I think this is fine.

src/ansys/grantami/bomanalytics/_typing.py
```diff
@@ -0,0 +1,50 @@
+from typing import Any, TypeVar, Union
+
+from ansys.openapi.common import Unset_Type
+
+Any_Type = TypeVar("Any_Type", bound=Any)
+
+
+def _cast_unset_union_to_any(value: Union[Any_Type, Unset_Type]) -> Union[Any_Type]:
```
ludovicsteinbach, line 8: Just trying to see if I'm getting this right. 

`Any_Type` is required to indicate that the returned value is of the same type as the input value (if it wasn't Unset)? Sort of a Generic situation?

Can the return be `Any_Type` or is the Union actually required?

src/ansys/grantami/bomanalytics/_item_results.py
ludovicsteinbach: Is `_cast_*` the right terminology here?
I was initially confused as to what it would do, because I thought it worked like the builtin typing.cast

src/ansys/grantami/bomanalytics/_item_results.py
```diff
@@ -732,9 +745,10 @@ def __init__(
 
         for legislation in legislations:
             new_substances = [
-                self._create_impacted_substance(substance) for substance in legislation.impacted_substances
+                self._create_impacted_substance(substance)
+                for substance in _cast_unset_union_to_any(legislation.impacted_substances)
```
ludovicsteinbach, line 749: Github won't let me expand the code before so I can't really see what's what. I'll try to check locally.

Is the casting necessary here? I can't see how `legislation.impacted_substances` could be None without this block raising an error.
I suppose the question is, was the casting applied based on test failures, mypy errors, or everywhere?

src/ansys/grantami/bomanalytics/_typing.py
Andy-Grigg: Ah, Union is not needed. As you say, it's a generic function as you say, the Any_Type typevar just makes it function for any type.

I'll fix the union

src/ansys/grantami/bomanalytics/_item_results.py
Andy-Grigg: I suppose I've used 'cast' in the sense that it will convert the type of something that contains a 'unset' type to something that does not contain an 'unset' type. Typically if a type conversion cannot complete for some reason, then it will raise an exception.

In this case, it will raise an exception because it's not possible to cast the value `Unset` to something that's not of `Unset_Type`.

src/ansys/grantami/bomanalytics/_item_results.py
Andy-Grigg: Yes, that's correct. My rationale being that in previous releases it was either a list or `None`, and we previously didn't check it wasn't `None`, so it seemed safe to change it to what we have now.

That's my philosophy behind all these changes:

* If we previously didn't check that a thing was None, then pass it through the generic function that raises if it's Unset. At least now it will fail with an explicit error message, instead of a `None type is not indexable` or whatever we'd get
* If we previously did check that a thing was None, then pass it through the generic function that converts Unset to None. I've created an openapi-client-template PR that includes this conversion in the models themselves, which would make this generic function unnecessary

Andy-Grigg: https://github.com/ansys/openapi-client-template/pull/168

Andy-Grigg: The first one, basically the vast majority of fields are not required and not nullable, so they end up with this Unset union type. It would be nice to get it fixed in the server definition - I will add a ticket in Jira to this effect.

Andy-Grigg: And no, Unset really only makes sense for models used for patch requests. The problem is, we don't distinguish between what a model is used for (request/response/verb), and so we're kind of stuck with it. The additional parallel models is intended to be a way of reconciling that.

tests/inputs/examples.py
Andy-Grigg: There are still some examples that include implicit null values (i.e. yaml fields without values) for non-nullable fields. I've changed the json loads hook to remove these fields from the resulting dict instead of populating them, which means we get back to deserializing to `Unset` and checking that the logic to convert to `None` works as expected.

src/ansys/grantami/bomanalytics/_item_results.py
ludovicsteinbach: I'm not sure I understand the point of `_raise_if_unset`: if we were previously not checking if something was None (because the API always includes even though it does not document it as such), why do we now need to check that it's not Unset?
Especially since, if it is, we'll now raise a cryptic ValueError

Andy-Grigg: I thought it was better to explicitly raise a ValueError indicating that the value received from the server was wrong, as opposed to a TypeError because whatever we happened to try to do to the incorrect value wasn't possible based on its type,

Although I realize now looking at the code that I didn't actually put a message in with the ValueError. This should probably be something like:

`raise ValueError("Provided value cannot be Unset")`

Maybe passing the name of the field as an extra argument, so we can raise with a more informative message?

Andy-Grigg: I should also add, the reason we raise is to satisfy mypy. The alternative would be to cast to `T`, which I think will work.

ludovicsteinbach (approved):

src/ansys/grantami/bomanalytics/_typing.py
```diff
@@ -0,0 +1,49 @@
+from typing import Any, Optional, TypeVar, Union
+
+from ansys.openapi.common import Unset_Type
+
+T = TypeVar("T", bound=Any)
+
+
+def _raise_if_unset(value: Union[T, Unset_Type], property_name: Optional[str] = None) -> T:
```
ludovicsteinbach, line 8:
```suggestion
def _raise_if_unset(value: Union[T, Unset_Type]) -> T:
```

PR #465, opened 2024-04-04
Add substance amounts to compliance

### LINKED ISSUE #454: Expose substance amounts in compliance responses
### 📝 Description of the feature

CR-1496 is including substance amounts in Compliance API responses. We should include these in our response objects once a new build is available.

### 💡 Steps for implementing the feature

_No response_

### 🔗 Useful links and references

_No response_

Andy-Grigg:
Closes #454 

Adds a new property to the `SubstanceWithComplianceResult` class to display the percentage amount of the substance in the parent item, and include in the Material Compliance example to show how to access the substance amount.

Added checks to the mocked response tests. These results have different amounts in different test cases, so we can't define the expected values in a separate location. In checking this though, I realized some of the mocked tests weren't actually testing anything, more details below.


tests/test_mocked_responses/common.py
```diff
@@ -255,7 +255,7 @@ def check_substance_details(self):
         """There are only 4 distinct substances in the example responses. This method checks that the substance is
         one of those 4 substances."""
 
-        return (
+        assert (
```
Andy-Grigg, line 258: We never actually tested the result of this method

tests/test_mocked_responses/common.py
```diff
@@ -223,10 +223,10 @@ def check_reference(
         """Extend base implementation to support cas_number, ec_number, and chemical_name for substances."""
 
         return (
-            self.obj.cas_number == cas_number
-            and self.obj.ec_number == ec_number
-            and self.obj.chemical_name == chemical_name
-            and super().check_reference(record_history_identity, record_guid)
+            (self.obj.cas_number == cas_number and cas_number is not None)
```
Andy-Grigg, line 226: We don't need all of them to match, we just need at least one to match that isn't None. I think the previous configuration was resulting in false positives if nothing was provided.

Andy-Grigg:
CI will fail until we can release a new bomanalytics-openapi alpha

src/ansys/grantami/bomanalytics/_item_results.py
```diff
@@ -1237,10 +1238,21 @@ class SubstanceWithComplianceResult(ComplianceResultMixin, SubstanceReferenceWit
     """Retrieves an individual substance included as part of a compliance query result.
     This object includes two categories of attributes:
 
-      - The reference to the substance in Granta MI
-      - The compliance status of this substance, stored in a dictionary of one or more indicator objects
+    * The reference to the substance in Granta MI
```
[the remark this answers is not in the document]
Andy-Grigg: Maybe? I only noticed this because I looked at the docs. The current approach indents the bullets with a vertical line, which looks a bit odd I think

PR #468, opened 2024-04-05
Add missing versionadded directives

### LINKED ISSUE #464: Remove versionadded directives
https://github.com/ansys/grantami-bomanalytics/blob/0e283eda033a99e0fe6d001c859ca86a6cf96908/src/ansys/grantami/bomanalytics/indicators.py#L376

It's not been used for all additions and it doesn't make sense to keep it forever

Andy-Grigg:
Closes #464 

Adds versionadded directives for all additions to the API since the initial release.

General approach was to add them for entire classes if the class is new, or individual properties or attributes if not. Also added to API pages. I didn't add the directive for every BoM type, because we want to keep this somewhat auto-generatable, and adding versionadded directives here wouldn't be possible with that approach.

### DIFF
--- a/src/ansys/grantami/bomanalytics/_item_definitions.py
+++ b/src/ansys/grantami/bomanalytics/_item_definitions.py
@@ -385,8 +385,7 @@ class CoatingReferenceWithIdentifier(IdentifierMixin, CoatingReference):
 class ProcessReference(RecordReference):
     # Because of ProcessSummaryResult, this is publicly documented.
     # Extends RecordReference without changes, to re-define the class name, because it appears in the repr.
-    """
-    Represents a reference to a Process record.
+    """Represents a reference to a Process record.
 
     .. versionadded:: 2.0
     """
@@ -397,8 +396,7 @@ class ProcessReferenceWithIdentifiers(CommonIdentifiersMixin, ProcessReference):
 
 
 class TransportReference(RecordReference):
-    """
-    Represents a reference to a Transport record.
+    """Represents a reference to a Transport record.
 
     .. versionadded:: 2.0
     """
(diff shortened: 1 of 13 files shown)


src/ansys/grantami/bomanalytics/_item_definitions.py
```diff
@@ -385,15 +385,21 @@ class CoatingReferenceWithIdentifier(IdentifierMixin, CoatingReference):
 class ProcessReference(RecordReference):
     # Because of ProcessSummaryResult, this is publicly documented.
     # Extends RecordReference without changes, to re-define the class name, because it appears in the repr.
-    """Represents a reference to a Process record."""
+    """.. versionadded:: 2.0
+
+    Represents a reference to a Process record.
+    """
```
ludovicsteinbach, line 391: I think I'd prefer if we kept the docstring first, it's a lot easier to read when looking at the source. Would it look decent in docs if the directive was at the end of the docstring?

Andy-Grigg: Updated the docstring to move the directives to the bottom

Andy-Grigg: image

src/ansys/grantami/bomanalytics/_item_results.py
```diff
@@ -1613,12 +1621,15 @@ class MaterialWithSustainabilityResult(
     MassResultMixin,
     MaterialReferenceWithIdentifiers,
 ):
-    """Describes an individual material included as part of a sustainability query result.
+    """
+    Describes an individual material included as part of a sustainability query result.
```
ludovicsteinbach, line 1625: Only flagging as I think it renders differently without the empty line. There are a couple more instances below

ludovicsteinbach (approved):

PR #513, opened 2024-05-06
Update openapi-common reference link

ludovicsteinbach:
Update openapi-common intersphinx link to accomodate change in object inventory location. (caused by new multiversioned docs)


ludovicsteinbach:
Merging despite the vale error: it's on a changelog fragment that has nothing to do with this PR

PR #517, opened 2024-05-06
Fix cross-references to openapi-common

### LINKED ISSUE #516: Fix links in documentation to openapi-common
We recently migrated the openapi-common documentation to a more PyAnsys-typical multi-version arrangement. Unfortunately, this has broken all of our cross-references between the documentation in this repo and the openapi-common documentation.

We could backport the fix from #513 to the other release branches and redeploy the documentation, but this will be very difficult because the underlying Granta MI API has changed, and so both the CI and example builds will fail.

Instead, we should manually fix the documentation on the docs branch in this repo. Any subsequent patch releases on old branches will fail the CI for reasons described above, so we don't have to worry about accidentally reverting these manual changes.

Andy-Grigg:
Closes #516 

Replace `https://openapi.docs.pyansys.com/` with `https://openapi.docs.pyansys.com/version/stable/api`.

I noticed that the link to the general 'API documentation' includes a version number in the `title` attribute, which I believe is what the version happened to be when the documentation was built. I believe this will continue, even referring to the 'stable' version of the intersphinx mapping file, I don't think there's much we can really do about this.

Fairly certain this catches them all.

### DIFF
--- a/version/dev/api/common.html
+++ b/version/dev/api/common.html
@@ -536,7 +536,7 @@ class instances of the <a class="reference internal" href="#ansys.grantami.boman
 </dl>
 <p class="rubric">Notes</p>
 <p>For advanced usage, including configuring session-specific properties and timeouts, see the
-<a class="reference external" href="https://openapi.docs.pyansys.com/version/stable/api/index.html" title="(in ansys.openapi.common stable)"><span class="xref std std-doc">ansys-openapi-common API reference</span></a>. Specifically, see the
+<a class="reference external" href="https://openapi.docs.pyansys.com/version/stable/api/index.html" title="(in ansys.openapi.common v2.0.2)"><span class="xref std std-doc">ansys-openapi-common API reference</span></a>. Specifically, see the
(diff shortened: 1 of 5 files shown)


version/dev/api/common.html
```diff
@@ -536,7 +536,7 @@ <h2>Connection builder<a class="headerlink" href="#connection-builder" title="Li
 </dl>
 <p class="rubric">Notes</p>
 <p>For advanced usage, including configuring session-specific properties and timeouts, see the
-<a class="reference external" href="https://openapi.docs.pyansys.com/version/stable/api/index.html" title="(in ansys.openapi.common v2.0.2)"><span class="xref std std-doc">ansys-openapi-common API reference</span></a>. Specifically, see the
```
ludovicsteinbach, line 539: Is this deliberate one? I think it would be overwritten on any pushes to main

ludovicsteinbach (approved):

version/dev/api/common.html
Andy-Grigg: I changed all these individual versions to 'stable', but realized that it wasn't something we'd be able to maintain, and so I changed them back. I must have missed this one, I'll revert before I merge.

Andy-Grigg:
Merging despite CI failures - CI is not intended to run on merges to this branch

PR #519, opened 2024-05-06
Do not run changelog-fragment for dependabot-initiated PRs

### LINKED ISSUE #515: The changelog PR workflow 'freezes' dependabot PRs
There is a subtle but irritating interaction between dependabot PRs and the changelog PR action. The following sequence of events occurs for dependabot PRs:

1. Dependabot opens a PR for a dependency change
2. The changelog action creates a changelog entry for the PR, and commits it
3. Dependabot sees that 'someone' else has modified the PR, and will then no longer update it

This means dependabot will not update the PR if another update happens. This could be potentially nasty if dependency groups are used, because the breadth of impact of a 'frozen PR' would extend, I believe, across the entire group, not just over a single dependency.

I think we should have the changelog action **not** trigger from dependabot updates. In the vast majority of cases getting a dependabot PR for every tiny dependency change is not useful - it's only useful if some kind of human intervention is needed to fix an incompatibility, and in this case a manually-opened PR will still trigger the changelog workflow.

Andy-Grigg:
Closes #515 

Don't run changelog-fragment for dependabot-initiated PRs to avoid dependabot PRs being 'frozen' by someone other than dependabot modifying the branch.

CI on this branch won't be affected, but I'm opening this PR for discussion. @klmcadams I've added you since you've been involved in some of our adventures with the new changelog actions.

### DIFF
--- a/.github/workflows/label.yml
+++ b/.github/workflows/label.yml
@@ -49,6 +49,7 @@ jobs:
         contents: write
         pull-requests: write
       runs-on: ubuntu-latest
+      if: ${{ !startsWith( github.event.pull_request.head.ref, 'dependabot/') }}
       steps:
       - uses: ansys/actions/doc-changelog@v6
         with:
--- a/doc/changelog.d/519.changed.md
+++ b/doc/changelog.d/519.changed.md
@@ -1,0 +1,1 @@
+Do not run changelog-fragment for dependabot-initiated PRs


klmcadams:
@Andy-Grigg This makes sense to me! Are your dependabot PRs automatically approved unless there's another author that commits to the branch?

Andy-Grigg:
> @Andy-Grigg This makes sense to me! Are your dependabot PRs automatically approved unless there's another author that commits to the branch?

It triggers as soon as the PR is created to just mark it as approved, so another user wouldn't have had change to commit. It's really just there to cut down on manual clicking, essentially so dependabot PRs can be merged. If CI on the PR fails this should still be resolved before merging (as enforced by the branch protection rules).

If there was manual intervention required, we'd generally close the dependabot PR and create a new one, or just leave it open so it gets closed by dependabot automatically, see https://github.com/ansys/openapi-client-template/pull/177 for a recent example

Andy-Grigg:
@da1910 Since Ludovic's gone on vacation, can you weigh in on this?

da1910 (approved):
This looks reasonable, it's a pain that we'll lose it from the changelog, I wonder if we can write a little script that makes an overall changelog entry for all dependencies in a release (if we decide we care).

da1910:
Although we probably want to delete the entry this PR created in that case

PR #526, opened 2024-05-07
Prepare 2.1 Release

da1910:
- [x] Update to release version of openapi package
- [x] Update version to release


Andy-Grigg:
I have updated the reports to the latest version on the test VM, but still transferring the MI installer.

Let's let CI run, at least that will check the latest reports definitely work. I won't approve though until we have a green run against latest MI.

Andy-Grigg (approved):

PR #546, opened 2024-05-31
Empty elements are no longer removed when deserializing a BoM file

Andy-Grigg:
Closes #545 

Standardizes the options used when deserializing a BoM from a file and from a string. Previously, `keep_empty=True` was only specified when reading a BoM from string, which caused a weird asymmetry in some situations.

This PR removes this duplication, and adds some tests that check deserializing and round-tripping a BoM with an empty element works both via a file and via text.

PR #478, opened 2024-04-10
CI - Check for active runs before stopping VM

ludovicsteinbach:

PR #547, opened 2024-06-03
Prepare 2.1.1 release

Andy-Grigg:
Cherry-pick the following PRs:

* #478
* #498  
* #513 
* #546

Update the version number.

~~I also want to include #510, but on merging the cherry-pick, it looks like the release job won't work correctly. I'll fix in main and then cherry-pick the fixed commit.~~ - Keeping the changelog fragments means upgrading the CI, which isn't possible to cherry-pick without pulling in a lot of documentation changes. Instead, I'll just delete the fragments and update the changelog by hand.


ludovicsteinbach (changes requested):
This should probably cherry-pick #498 as well

Andy-Grigg:
Codacy seems to be failing to compute diffs and deltas for this PR, but there's no meaningful information as to what's going wrong.

PR #607, opened 2024-09-11
Update acceptance tests for 2025 R1 database

Andy-Grigg:
Update the hard-coded acceptance test values to align with updated data in the 2025 R1 database.

Also standardize some whitespace and item namings in the test XML BoM.

There is a change to the number of primary processes reported by the BoM Analytics Service, even though the BoM hasn't changed. I believe this is because of the adjusted numbers, and that now this process has just tipped above the threshold to be considered a result in its own right, instead of being lumped in with 'other'.

### DIFF
--- a/examples/4_Sustainability/supporting-files/bom-2301-assembly.xml
+++ b/examples/4_Sustainability/supporting-files/bom-2301-assembly.xml
@@ -32,7 +32,7 @@
 											<Percentage>100.0</Percentage>
 										</Process>
 										<Process>
-											<MIProcessReference >
+											<MIProcessReference>
 												<dbKey xmlns="http://www.grantadesign.com/12/05/GrantaBaseTypes">MI_Restricted_Substances</dbKey>
 												<recordGUID xmlns="http://www.grantadesign.com/12/05/GrantaBaseTypes">907bda29-e800-44f6-b7ea-4eb8e7cff375</recordGUID>
 											</MIProcessReference>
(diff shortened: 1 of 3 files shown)


examples/4_Sustainability/supporting-files/bom-2301-assembly.xml
```diff
@@ -120,7 +120,7 @@
 										<dbKey xmlns="http://www.grantadesign.com/12/05/GrantaBaseTypes">MI_Restricted_Substances</dbKey>
 										<recordGUID xmlns="http://www.grantadesign.com/12/05/GrantaBaseTypes">92f90382-610a-4737-a18f-b153ea88e4ad</recordGUID>
 									</MIProcessReference>
-									<Identity>Primary processing, Metal extrusion, hot</Identity>
+									<Identity>Metal extrusion, hot</Identity>
```
ludovicsteinbach, line 123: This could the source of the additional item in the response. IIRC the process identity is part of the aggregation criteria, and there already are two instances of this process applied to this material, so as you said, it now meets the threshold.

What's the reason for changing this? (it's not unreasonable, just curious how that showed up)

ludovicsteinbach (approved):

examples/4_Sustainability/supporting-files/bom-2301-assembly.xml
Andy-Grigg: Oh, I didn't realize that was the case. I changed it because I thought it was just an inconsistency. I'll try reverting it and I'll see what happens.

ludovicsteinbach: It might well be an inconsistency, I can't remember if it's intentionally different

Andy-Grigg: I just tried reverting back to the `main` version of this file, and it didn't make a difference to the test result.

ludovicsteinbach: I checked the docs, aggregation is based on process name and material identity, so this change should have no effect on the summary calculation. (my bad)

Andy-Grigg: No worries. In that case, it really must be that it was marginally below the threshold before and has now gone above the threshold.

I've asked Rachel to confirm that changes have been made to the database that would cause changes to the results, but aside from that I don't think we need to go into any more detail.

PR #635, opened 2024-10-25
Upgrade to grantami-bomanalytics-openapi 3.1.0rc1

da1910:
Additional changes:

- Support 3.13
- Drop support for 3.9
- Switch to 3.12 for main python in CI


.github/workflows/ci_cd.yml
```diff
@@ -10,7 +10,7 @@ on:
   merge_group:
 
 env:
-  MAIN_PYTHON_VERSION: '3.9'
+  MAIN_PYTHON_VERSION: '3.12'
```
Andy-Grigg, line 13: Other pyansys packages tend to use the oldest supported Python package for this, as far as I can tell. We should be consistent.

```suggestion
  MAIN_PYTHON_VERSION: '3.10'
```

.github/workflows/server_checks.yml
```diff
@@ -4,7 +4,7 @@ on:
       python-version:
         description: >
           Python version used for installing and running ``pre-commit``.
-        default: '3.10'
+        default: '3.12'
```
Andy-Grigg, line 7:
```suggestion
        default: '3.10'
```

.github/workflows/ci_cd.yml
da1910: Hmm, I made this change on all our packages...

da1910: https://github.com/ansys/grantami-recordlists/blob/main/.github/workflows/ci_cd.yml
https://github.com/ansys/grantami-jobqueue/blob/main/.github/workflows/ci_cd.yml
https://github.com/ansys/grantami-serverapi-openapi/blob/main/.github/workflows/build_and_test_library.yml
https://github.com/ansys/grantami-bomanalytics-openapi/blob/main/.github/workflows/build_and_test_library.yml

Andy-Grigg: I had a look at some other repos, e.g. [pymapdl](https://github.com/ansys/pymapdl/blob/ededf7320da0594b4b050ddeec630666c403069e/.github/workflows/ci.yml#L23), and there they're using 3.11. Maybe there is no convention, in which case this is fine.

Andy-Grigg (approved):

PR #650, opened 2024-11-15
Prepare for 2.2.0rc0 release

Andy-Grigg:
Prepare for 2.2.0rc0 release

### DIFF
--- a/doc/changelog.d/650.dependencies.md
+++ b/doc/changelog.d/650.dependencies.md
@@ -1,0 +1,1 @@
+Prepare for 2.2.0rc0 release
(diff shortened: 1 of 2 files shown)


Andy-Grigg:
@ludovicsteinbach  You might be fighting a losing battle with the label. I shouldn't have changed the lockfile anyway, I'll force push this branch without that change

Andy-Grigg:
Codacy failure seems to be a false negative

ludovicsteinbach (approved):
The changelog fragment is still incorrectly categorized as 'dependencies' if that's something we care about.
Other than that this looks reasonable

Andy-Grigg:
Good point. I wonder if I can fix it by deleting the commit.

ludovicsteinbach:
Although the openapi dependency should probably not accept pre releases anymore?

Andy-Grigg:
> Although the openapi dependency should probably not accept pre-releases anymore?

Changing that here was the cause of the pain, because it required a change to the lockfile.  The dependabot PR didn't change it.

Andy-Grigg:
This PR seems borked. Recreating.

Andy-Grigg:
Recreated here #652

PR #669, opened 2024-12-13
Fix BoM Analytics Services dependency

Andy-Grigg:
Update the BoM Analytics Services dependency to 2024 R2, which is required because of https://github.com/ansys/grantami-bomanalytics/pull/465.

Also clarify the source of the dependency in the metapackage instructions, as being dependent on BAS, as opposed to Granta MI itself.

This fix should be cherry-picked onto the release/2.2 and release/2.1 branches following merge.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -52,9 +52,8 @@ Dependencies
 ------------
 .. readme_software_requirements
 
-To use the ``grantami-bomanalytics`` package, you must have access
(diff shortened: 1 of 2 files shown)


ludovicsteinbach (commented):
Should the run-time check version also be 24R2?

ludovicsteinbach (approved):

Andy-Grigg:
> Should the run-time check version also be 24R2?

Addressed in #670

PR #693, opened 2025-01-31
Support 24/12 Eco BoM

### LINKED ISSUE #689: Enhance BoM Handler to support 2412 XSD
The new 2412 XSD schema for bills of materials supports transport and location child nodes for additional node types. The builder needs to be enhanced to allow these new features.

We should consider as well if we want to continue to support the current 2301 format. A few thoughts on this:

1. Completely duplicate all the BoM building code and leave the old version as-is
2. Use different methods, or arguments in a method, to control which version we serialize to

Second option is better, but we'd need to handle how to serialize objects that cannot be represented in older BoM formats. Maybe raise an exception if these are encountered? Or delete them and warn?

Either way, we need to make sure this is an additive change, so scripts written against the previous version of the BoM handler continue to function. This probably means option two, using an argument to determine the BoM output. We should be able to default to the latest version, and offer an override if needed.

Andy-Grigg:
Closes #689 

* Add a new `eco2412` submodule for the 24/12 eco schema.
* Refactor the `bom_helper` submodule to be able to handle different BoMs with different namespaces. Any supported BoM can be loaded and dumped, and a BoM can be converted to a different namespace

This is part of a series of 3 PRs:

1. #702 
2. #703
3. This PR


src/ansys/grantami/bomanalytics/bom_types/__init__.py
```diff
@@ -20,7 +20,5 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-from ._bom_reader import BoMReader
-from ._bom_types import *
-from ._bom_writer import BoMWriter
-from ._builders import *
+from .eco2412 import *
```
Andy-Grigg, line 23: Assuming it's fine that eco2412 becomes the new default BoM format when creating new BoMs. This does mean that people creating a bom from scratch and are using an older version of the reports bundle will need to downgrade their BoM first, or need to change their import.

This is the same grey area with semantic version where not upgrading the server causes an issue. Even though the 24/12 schema is fully backwards compatible with the 23/01 schema, they're still different types.

Needs more thinking as to whether this constitutes a major version, or if it can be handled server side.

Alternatively, we could expose these types as the default, but always try and downgrade the bom when dumping, unless a more modern format is specified explicitly.

ludovicsteinbach: I think the way we implemented support for multiple BoM versions on the endpoints means that it's unlikely to break if we change the default BoM type.
But, we don't know that this is only thing users would be doing after generating a BoM with the package and they might well have some internal other operations that relies on the BoM schema not changing.

We could consider implementing the new format but keeping the 23/01 as the default until we truly need to make a new major release?

Andy-Grigg: I think you're right, and my original commit had `from .eco2301 import *`. I changed it because I wanted the 2412 stuff to be discoverable, since what we don't want is someone trying to use the new functionality and being unable to because they didn't import things correctly. I also went in this direction because the 2412 schema is fully backwards compatible with 2301, and assuming we can figure out what schema version is supported with which API version, we can make it work.

However, that's potentially a lot of messing around, and as you say, the entire point of the bomhandler stuff is that you don't necessarily need to be doing BoM Analytics stuff with it. And since XML namespaces are not backwards compatible, it doesn't matter if one schema happens to be a subset of another, they're still essentially incompatible.

Ideally, we'd have foreseen this eventually and would have created a separate submodule for `eco2301` with the original 2.0 release. We didn't do that, and really we had no grounds to predict this is where we'd end up. Ultimately though, I think in the next major release we're better removing the imports into the `bom_types` submodule altogether, so you _have_ to pick a version explicitly.

ludovicsteinbach: > I think the way we implemented support for multiple BoM versions on the endpoints means that it's unlikely to break if we change the default BoM type. 

That would be true if there were endpoints for analysis on 24/01 BoMs, but that apparently hasn't been added to the API.
Is it only on the v2? If I remember correctly v2 was meant to have BoM endpoints accept multiple BoM formats rather than having multiple endpoints each supporting a single BoM format.

Andy-Grigg: Exactly. v2 has schema-agnostic endpoints, so we'd need to support v2 for this new analysis. However, all supported versions of MI don't support v2. Since they're separate API definitions, we need to support multiple bindings in the same bomanalytics-openapi package if we want to continue to support older versions.

src/ansys/grantami/bomanalytics/bom_types/_bom_reader.py
```diff
@@ -43,10 +43,25 @@ def __init__(self) -> None:
...
         accepts a schema, and _class_members property should be set to classes to deserialize to.
         """
+
         self._namespaces: Dict[str, str] = {}
+
         # Used to track fields in an object that haven't been deserialized.
         self.__undeserialized_fields: list[str] = []
 
+    @property
+    def eco_namespace(self) -> str | None:
```
ludovicsteinbach, line 53: `eco_namespace` is surprising, since `eco` is really just an arbitrary string used as prefix.
Although I realise now that has been introduced as a concept when the existing bom types were moved to a `eco2301` sub-package. 
This might be a light case of semantic satiation on my end, but it doesn't feel like to `eco` is the best choice here.

src/ansys/grantami/bomanalytics/bom_types/eco2412/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from ._bom_reader import BoMReader
+from ._bom_types import *
+from ._bom_writer import BoMWriter
+
+__all__ = [
+    "BillOfMaterials",
```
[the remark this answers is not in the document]
Andy-Grigg: I had to re-work this from the previous PR. In that case I said that I just wouldn't include them in the `__init__.py` and you can get them from the module instead. However, that becomes a pain here because we just want to import the XML namespace module (i.e. `eco2301` or `eco2412`) and access the relevant classes from there.

You can't do that if you don't import the actual module that includes the class, and it just gets complicated with additional dictionaries having to exist to do the mapping.

I've gone back to making them private classes and importing them into the top-level module. I did wonder if we need `__all__` at all, but it's useful for documentation, because it controls the order in which automodule outputs the module classes.

src/ansys/grantami/bomanalytics/bom_types/_bom_reader.py
Andy-Grigg: This is the part of the whole implementation I'm most concerned about, and it feels like the thing that could go wrong if someone does something legal but unexpected with a Bill of Materials.

I've looked into this a bit, and I think what I actually should be referring to is `targetNamespace`. This is a well-defined concept in an XSD: https://stackoverflow.com/questions/7190572/targetnamespace-and-xmlns-without-prefix-what-is-the-difference

This is what I think is needed. We want to know the target namespace for the reader/writer, and match it up with the namespace of the BillOfMaterials object in the BoM we want to serialize/deserialize.

Andy-Grigg:
The first example test sometimes times out. I'll try upgrading, see if that fixes it.

src/ansys/grantami/bomanalytics/bom_types/_bom_reader.py
```diff
@@ -24,16 +24,19 @@
...
+from xmlschema import XMLSchema
+
 from ._base_types import BaseType, HasNamespace
 
 TBom = TypeVar("TBom", bound=BaseType)
 TAny = TypeVar("TAny", bound=BaseType)
 
 
-class GenericBoMReader(Generic[TBom]):
+class _GenericBoMReader(Generic[TBom]):
```
Andy-Grigg, line 35: Renaming these to be private methods means that they can be imported into `__init__.py`, which makes them easier to import elsewhere.

Andy-Grigg:
Since we've converted the reader and writer classes to generics, we could define the BoMHandler class itself as a generic instead.

It's a choice between whether we want to make the user decide ahead of time which version BoM version they want to work with or not.

doc/source/api/bom_builder/eco2301.rst
```diff
@@ -5,12 +5,16 @@
 
 .. currentmodule:: ansys.grantami.bomanalytics.bom_types.eco2301
 
+.. warning::
```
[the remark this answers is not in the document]
Andy-Grigg: That's a good point, I was probably too enthusiastic with this comment.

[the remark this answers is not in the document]
Andy-Grigg: I suppose we don't support reports earlier than 2024 R1 at all, so we don't need to state it. We should state the minimum report version for the 24/12 BoM schema though.

src/ansys/grantami/bomanalytics/_bom_helper.py
```diff
@@ -19,38 +19,70 @@
...
 
 if TYPE_CHECKING:
-    from .bom_types import BillOfMaterials
+    from .bom_types import _GenericBoMReader, _GenericBoMWriter
+    from .bom_types.eco2301 import BillOfMaterials as BillOfMaterials2301
+    from .bom_types.eco2412 import BillOfMaterials as BillOfMaterials2412
+
+
+BillOfMaterials: TypeAlias = "BillOfMaterials2301 | BillOfMaterials2412"
+T = TypeVar("T", bound=BillOfMaterials)
```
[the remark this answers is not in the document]
Andy-Grigg: I suppose not. I did this though because Python won't let you have a single constraint, even if that constraint is a union of types, because you could just replace that typevar with the type itself. I'm not entirely sure that argument holds here because we want to know which of the two types the user selected, but anyway, that's the issue.

I can just provide each type explicitly, which as you say, narrows the scope of `T`.

tests/test_bom_handler.py
```diff
@@ -530,15 +636,107 @@ def make_full_bom(self, use_phase_utility_kwarg) -> bom_types.BillOfMaterials:
...
+
+@pytest.mark.parametrize(
+    "use_phase_utility_kwarg",
+    [
+        "industry_average_number_of_functional_units",
+        "industry_average_duration_years",
+        "utility",
+    ],
+)
+class TestBoMConversion:
```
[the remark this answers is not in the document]
Andy-Grigg: Classic example of what happens if you don't take a step back and clean up what you've written. I think splitting the tests up actually reduced the number of lines of code.

ludovicsteinbach (approved):

PR #702, opened 2025-02-12
Optionally raise exceptions if a BoM can't be fully deserialized

Andy-Grigg:
Add an additional argument to BOM load methods `load_bom_from_file` and `load_bom_from_text` to enforce 'strict' deserialization.

The new argument `allow_unsupported_data` defaults to `True`, which follows the current behavior of allowing data which is compliant with the BoM schema, but is not supported by the classes in the `bom_types` submodule.

This is to allow control when downgrading BoMs as part of the 24/12 work. The downgrading will be implemented as a separate method, and that method will also include this argument. In that case though, the unsupported classes will more likely be caused by more modern BoMs that cannot be fully represented in older versions.

The first in a series of three PRs:

1. This PR
2. https://github.com/ansys/grantami-bomanalytics/pull/703
3. https://github.com/ansys/grantami-bomanalytics/pull/693

### DIFF
--- a/src/ansys/grantami/bomanalytics/_bom_helper.py
+++ b/src/ansys/grantami/bomanalytics/_bom_helper.py
@@ -69,10 +69,11 @@ class BoMHandler:
         Raises
         ------
         ValueError
-            If the BoM cannot be deserialized.
+            If the BoM cannot be deserialized. Additional detail is included in the exception message.
         ValueError
             If the BoM contains data that cannot be represented by the classes in the
-            :ref:`ref_grantami_bomanalytics_bom_api` and ``allow_unsupported_data = False`` is specified.
+            :ref:`ref_grantami_bomanalytics_bom_api` and ``allow_unsupported_data = False`` is specified. The additional
+            data is reported in the exception message.
         """
         with open(file_path, "r", encoding="utf8") as fp:
             obj, errors = self._deserialize_bom(fp)
@@ -86,7 +87,7 @@ class BoMHandler:
         bom, undeserialized_fields = self._reader.read_bom(obj)
         if undeserialized_fields and not allow_unsupported_data:
             self._raise_undeserialized_fields(undeserialized_fields)
-        return cast("BillOfMaterials", bom)
+        return bom
 
     def load_bom_from_text(self, bom_text: str, allow_unsupported_data: bool = True) -> "BillOfMaterials":
         """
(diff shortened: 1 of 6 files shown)


src/ansys/grantami/bomanalytics/_bom_helper.py
```diff
@@ -49,18 +49,30 @@ def __init__(self) -> None:
...
         -------
         :class:`~._bom_types.BillOfMaterials`
+
+        Raises
+        ------
+        ValueError
+            If the BoM cannot be deserialized.
+        ValueError
+            If the BoM contains data that cannot be represented by the classes in the
+            :ref:`ref_grantami_bomanalytics_bom_api` and ``allow_unsupported_data = False`` is specified.
```
ludovicsteinbach, line 75: It looks a bit weird to list the same error twice. Is the first just a very generic failure, for example if the provided file isn't xml at all or missing a closing `>`?

src/ansys/grantami/bomanalytics/bom_types/_bom_reader.py
```diff
@@ -50,8 +50,10 @@ def __init__(self, schema: XMLSchema):
         self._class_members: Dict[str, Type[BaseType]] = {
             k: v for k, v in inspect.getmembers(bom_types, inspect.isclass)
         }
+        # Used to track fields in an object that haven't been deserialized.
+        self.__undeserialized_fields: list[str] = []
```
ludovicsteinbach, line 54: Any particular reason for using a double underscore?

ludovicsteinbach (approved):

src/ansys/grantami/bomanalytics/_bom_helper.py
Andy-Grigg: It's if we get an error back from `XMLSchema.decode`. The BoM at this point is just a string, so it's if anything goes wrong from turning the string into the nested dictionary representation. Could be invalid XML, or could be an invalid namespace. I could add a note in here saying something like "The error message will contain specific information on the failure" or something.

src/ansys/grantami/bomanalytics/bom_types/_bom_reader.py
Andy-Grigg: This is an artifact of splitting up the big PR. By the end of this chain of PRs, this code is subclassed. I wanted to make sure the attribute name was mangled in case it was accidentally overriden.

PR #703, opened 2025-02-12
Move bom types to submodule

Andy-Grigg:
Refactoring PR to move the existing `bom_types` classes into submodules that represent the XML namespace they belong to. This will allow us to introduce new eco namespaces more easily.

It's a shame that we didn't do this earlier, since for backwards compatibility reasons the 23/01 classes are now available both in the root `bom_types` module and in the `bom_types.eco2301` module as well. This may be something that we can choose to change in a future major release.

Documentation is a bit awkward. I've done some minor re-organization, but it might be that we should overhaul the documentation anyway, and go with an approach which much closer represents the structure of the package itself.

The second in a series of 3 PRs:

1. #702 
2. This PR
3. https://github.com/ansys/grantami-bomanalytics/pull/693


Andy-Grigg:
codecov complains all seem to be false positives, and relate to TYPE_CHECKING imports or moved code which cannot be executed.

src/ansys/grantami/bomanalytics/bom_types/eco2301/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+from ._bom_reader import BoMReader
+from ._bom_types import *
+from ._bom_writer import BoMWriter
+
+__all__ = [
```
ludovicsteinbach, line 27: Should this include BoMReader and BoMWriter?

src/ansys/grantami/bomanalytics/bom_types/eco2301/_bom_types.py
```diff
@@ -24,99 +24,18 @@
...
-
-        Parameters
-        ----------
-        obj: Dict
-            Dictionary representing the current state of the serialization of self. Modified in place by this method.
-        bom_writer: BoMWriter
-            Helper object that maintains information about the global namespaces.
-        """
+class BaseType2301(BaseType):
+    namespace = "http://www.grantadesign.com/23/01/BillOfMaterialsEco"
```
ludovicsteinbach, line 38: Is making this attribute public done to actually make it part of the public API or because it's required for some other class to read?

src/ansys/grantami/bomanalytics/bom_types/gbt1205/_bom_types.py
```diff
@@ -0,0 +1,264 @@
...
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
```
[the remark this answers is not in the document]
Andy-Grigg: It was mainly to avoid duplication. The `grantarecord1205.xsd` document and associated `gbt`-prefix namespace is a logically separate entity, so it should live in its own location.

I don't think we need to worry about gbt changing though, it's already 13 years old and describes very fundamental aspects of the database.

src/ansys/grantami/bomanalytics/bom_types/eco2301/__init__.py
Andy-Grigg: `__all__` is what you get if you do `from eco2301 import *`. I thought it would be best to not include these here, because they're not types that a user is ever going to want to use, and not types that we document. I suppose if they're not included in `__all__`, they should probably be marked private as well.

src/ansys/grantami/bomanalytics/bom_types/eco2301/_bom_types.py
Andy-Grigg: It's mainly done because it means once we support multiple BoM versions, the bom handler can look into the `BillOfMaterials` object to see which writer it needs to use to serialize the BoM without just trying them all. So, because it's required for other classes to read.

I don't see any harm in making it part of the public API though, the namespace of an XML type is a pretty sensible thing to want to know.

src/ansys/grantami/bomanalytics/bom_types/_bom_reader.py
```diff
@@ -20,36 +20,30 @@
...
-from . import _bom_types as bom_types
-from ._bom_types import BaseType, HasNamespace
+from ._base_types import BaseType, HasNamespace
 
 if TYPE_CHECKING:
     from . import BillOfMaterials
 
 
-class BoMReader:
+class GenericBoMReader(ABC):
```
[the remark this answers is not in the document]
Andy-Grigg: I don't know this would be done. The input to the `read_bom` method (which is the thing that needs to have a generic return type) only as a dictionary as the parameter type. I would like to do something nicer here though, an explicit cast in the bom handler is a bit nasty.

src/ansys/grantami/bomanalytics/bom_types/eco2301/_bom_types.py
ludovicsteinbach: There is no harm indeed, it's a personal dislike.
I often find myself annoyed at our openapi autogenerated code because models have public properties that are actually relevant to what they are, but also a bunch of public properties that "pollute", e.g. `swagger_types` & `attribute_map` which are added by the base model.
But this is certainly the right way to do this.

src/ansys/grantami/bomanalytics/bom_types/eco2301/__init__.py
Andy-Grigg: I suppose taking that further, why are we bothering importing it into the top-level module at all.

src/ansys/grantami/bomanalytics/bom_types/_bom_reader.py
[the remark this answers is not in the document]
Andy-Grigg: I think the best I can do here is probably do the cast in the subclass. The base method returns a `BaseType`, and then the equivalent method in the subclass casts the response to the actual BillOfMaterials object for that submodule.

src/ansys/grantami/bomanalytics/bom_types/eco2301/__init__.py
Andy-Grigg: I've just removed the imports entirely, and we can get to them by referencing their respective modules directly.

src/ansys/grantami/bomanalytics/bom_types/_bom_reader.py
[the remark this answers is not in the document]
Andy-Grigg: I've renamed them to BaseBomreader and BaseBomWriter as well, to avoid making any promises about them being generic functions/classes.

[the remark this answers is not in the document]
ludovicsteinbach: My initial comment originates from changing the return type of `BoMReader.read_bom` from a specific `BillOfMaterials` type to any class derived from `BaseType`, because that's a bit too loose for my taste, even though it's correct.

Prior to these changes it was easier to associate the BoMReader and BoMWriter to a specific BoM type because there was only one. These base classes now implement behavior common to multiple derived classes, and I'd prefer to see these derived classes properly define which BoM they work with.

It's a bit difficult to detail in a comment, but I had a crack at it on a separate branch https://github.com/ansys/grantami-bomanalytics/tree/bom-types-typing
It does require making a few more changes, maybe that's not worth the trouble !

src/ansys/grantami/bomanalytics/bom_types/eco2301/__init__.py
ludovicsteinbach: Seeing as the old `bom_types.__init__` did not define `__all__` I think we'll have to include them at least in the top-level to ensure backwards compatibility.
Agreed that there is no point in including them in the new sub-packages `__all__`

src/ansys/grantami/bomanalytics/bom_types/_builders.py
```diff
@@ -24,7 +24,12 @@
...
-from . import MIAttributeReference, MIRecordReference, PartialTableReference, PseudoAttribute
+from .gbt1205._bom_types import (
+    MIAttributeReference,
+    MIRecordReference,
+    PartialTableReference,
+    PseudoAttribute,
+)
 
 
 class _AttributeReferenceByNameBuilder:
```
ludovicsteinbach, line 35: These builders are very much specific to the gbt1205 types, should they be moved to the sub-package?

src/ansys/grantami/bomanalytics/bom_types/eco2301/__init__.py
Andy-Grigg: Argh, you're right. They were public classes and were included in the bom_types `__init__.py`.

We didn't document them, so it's highly unlikely they're being used, but a public interface is a public interface. I'll adjust.

src/ansys/grantami/bomanalytics/bom_types/_builders.py
Andy-Grigg: I moved the builders out because in principle they could return any type. We could decide we want a material builder at some point that lets you add processes in a builder fashion. Right now they only build gbt1205 types, but there's no reason for that to be the case.

src/ansys/grantami/bomanalytics/bom_types/_bom_reader.py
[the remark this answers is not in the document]
Andy-Grigg: Yoink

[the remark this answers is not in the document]
Andy-Grigg: I should elaborate.

I've written generic functions and methods before, but I haven't worked with generic classes. This is what I was looking 
for though. Thank you for putting it together. I've changed the `Base` prefix to `Generic`, since that's now a more accurate description.

I just cherry-picked that commit onto this branch, and then cleaned up the naming.

src/ansys/grantami/bomanalytics/bom_types/_bom_reader.py
```diff
@@ -64,7 +59,7 @@ def read_bom(self, obj: Dict) -> tuple["BillOfMaterials", list]:
 
         Returns
         -------
-        tuple[BillOfMaterials, list]
+        tuple[T, list]
```
ludovicsteinbach, line 62:
```suggestion
        tuple[TBom, list]
```

src/ansys/grantami/bomanalytics/bom_types/_bom_reader.py
```diff
@@ -78,12 +73,17 @@ def read_bom(self, obj: Dict) -> tuple["BillOfMaterials", list]:
...
-        bom = cast("BillOfMaterials", self.create_type("BillOfMaterials", obj))
+        bom = self._create_type(self._bom_type, obj)
         return bom, self.__undeserialized_fields
 
     def create_type(self, type_name: str, obj: Dict) -> BaseType:
+        target_type = self._class_members[type_name]
+        return self._create_type(target_type, obj)
+
+    def _create_type(self, type_: Type[TAny], obj: Dict) -> TAny:
         """
```
ludovicsteinbach, line 84: I hadn't updated the docstring correctly: this is the docstring for the public `create_type` method

ludovicsteinbach (approved):

PR #710, opened 2025-02-21
Test on multiple servers

### LINKED ISSUE #712: Run CI against MI and Reports 24 R2, 25 R2, and dev
We support versions of the reports 24 R2 and later. We should run CI against deployments of all these versions, instead of just the latest.

Andy-Grigg:
Closes #712 

Add tests on multiple servers running different versions of Granta MI and RS & Sustainability reports bundle

* Adds references to additional servers in CI. Since secrets can't be added to a text matrix directly, the secret name is added as a string, and it is then later dereferenced.
* Add a parameter to the `integration` pytest mark, which is used to selectively skip tests based on MI/Reports version. This is mainly used for integration tests which are very specific about the results from the analysis.
* Reinstated the 24R2 version of the integration tests that have changed.

Currently only the servers that require testing for this branch have been added, but all pygranta test servers are shutdown.

Before this PR is merged, a similar change should be made to RecordLists and JobQueue, at least to shut down all servers on completion.

### DIFF
--- a/.github/workflows/server_checks.yml
+++ b/.github/workflows/server_checks.yml
@@ -146,7 +146,7 @@ jobs:
 
       - name: "Upload coverage artifacts"
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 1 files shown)


ludovicsteinbach:
In the context of BoM Analytics, the MI version for tests is also the reports version and the RS DB version?

Issue #690, opened 2025-01-29
Support new transport and location responses for sustainability analysis


Andy-Grigg:
The possible results from a sustainability analysis have been enhanced. Specifically:

* Parts and processes can now have a list of transport stages, which contain sustainability results
* Sustainability summaries now summarize transport by category
* Sustainability summaries now summarize transport by part

We should add these to the API and extend the examples to include these new cases.

Additive change, so shouldn't require a major version bump. These should be backwards compatible as well, in that they'll just be empty if the analysis is run against a previous version of the API.

Andy-Grigg:
I have made some progress with this, work is on a branch here https://github.com/ansys/grantami-bomanalytics/tree/feat/process-part-transport-location. A few things are starting to click into place:

* The enhancements should be backwards compatible, in that previous versions of the API will not use the new types, and will not accept BoMs of the new 24/12 format
* The variant of the integration tests for the latest version of the API should be extended to use the 24/12 format. The other integration tests can continue to use the 23/01 schema. I don't think it's necessary to run 23/01 tests against the latest BAS version.
* The examples will need to be upgraded to showcase the new functionality. This won't work against older versions of the API, so we need to:
  * Add a clear note to the relevant examples that the notebook as-written only supports a certain version. They can consult previous versions of the documentation for a compatible example
  * Add logic to conftest.py to skip these new examples on incompatible platforms.

Andy-Grigg:
> I don't think it's necessary to run 23/01 tests against the latest BAS version.

Turns out that this is trivial given how we parametrize tests

PR #714, opened 2025-02-24
Migrate to BoM Analytics Services V2

Andy-Grigg:
Migrate to BAS v2, which will include support for transport and location information on parts and processes.

Required for #690


src/ansys/grantami/bomanalytics/_query_results.py
```diff
@@ -766,26 +763,20 @@ class BomSustainabilityQueryResult(ResultBaseClass):
...
                 "Found no part in BoM sustainability response. Ensure the request BoM defines a single root part."
             )
-        if len(self._response.parts) > 1:
-            warnings.warn(
-                f"BomSustainabilityQuery only supports a single root part (found {len(self._response.parts)}). "
-                f"Additional root parts do not include sustainability results and are not exposed in the query result"
-                f" properties."
-            )
         # Exposing only a single root part:
         # API V1 only processes the first root part but still returns part empty part objects for extra root parts.
```
da1910, line 776: Do we still need this comment?

da1910 (approved):
Code changes look reasonable, one question about a comment

src/ansys/grantami/bomanalytics/_query_results.py
Andy-Grigg: Good point, probably not. V2 does what we would expect, so both comments can be removed.

Issue #687, opened 2025-01-28
Examples broken with plotly 6


Andy-Grigg:
### 🔍 Before submitting the issue

- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### 🐞 Description of the bug

I merged a dependabot PR which allowed the use of plotly v6. Unfortunately this seems to have broken the plotly plots in the examples:

https://bomanalytics.grantami.docs.pyansys.com/version/dev/examples/4_Sustainability/4-1_BoM_Sustainability_summary.html

We need to look at this to figure out if we can support v5 and v6, or if we need to change the code and use v6 only in the examples.

### 📝 Steps to reproduce

N/A - CI

### 💻 Which operating system are you using?

Linux

### 📀 Which ANSYS version are you using?

N/A

### 🐍 Which Python version are you using?

3.13

### 📦 Installed packages

```shell
Plotly v6
```

Andy-Grigg:
This seems to be an issue with plotly, and not something we can work around.

https://github.com/plotly/plotly.py/issues/5012

I see a very similar-looking error in our generated docs:

```
4-2_BoM_Sustainability.html:4777  Uncaught TypeError: kP.select is not a function
    at Ehe (4-2_BoM_Sustainability.html:4777:259206)
    at Object.Yft [as _doPlot] (4-2_BoM_Sustainability.html:4777:255283)
    at Object.Qft [as newPlot] (4-2_BoM_Sustainability.html:4777:260707)
    at 4-2_BoM_Sustainability.html:8635:371
Ehe @ 4-2_BoM_Sustainability.html:4777
Yft @ 4-2_BoM_Sustainability.html:4777
Qft @ 4-2_BoM_Sustainability.html:4777
(anonymous) @ 4-2_BoM_Sustainability.html:8635
```

Pinning to plotly<6 is probably the best solution for now.

PR #718, opened 2025-02-25
Xfail failing test

Andy-Grigg:


tests/conftest.py
```diff
@@ -96,12 +96,27 @@ def pytest_generate_tests(metafunc):
...
+
+    # Required to only xfail a test if we are running a specific version of MI
+    # The example_script fixture is only used for integration tests, and if we are running integration tests we must
+    # have an accessible instance of Granta MI
+    mi_version = _get_mi_version()
+
+    this_file = pathlib.Path(__file__).parent.resolve()
+    example_path = this_file / pathlib.Path("../examples")
+    output_files = discover_python_scripts(example_path)
+    param_sets = [
```
Andy-Grigg, line 110: This is all a bit indirect, but I'm not sure of an alternative given how we dynamically configure these tests.

We'll probably need something like this in future though, at least with the mi_version check, given that the current development work will introduce examples that can only be executed against the latest release of the BAS API.

tests/test_integration_tests.py
```diff
@@ -263,8 +263,9 @@ def _check_percentages_add_up(self, items):
         assert sum(item.embodied_energy_percentage for item in items) == pytest.approx(100)
         assert sum(item.climate_change_percentage for item in items) == pytest.approx(100)
 
-    @pytest.mark.integration(mi_versions=[(25, 1), (25, 2)])
-    def test_sustainability_summary_query_25_1(self, connection, mi_version, request):
+    @pytest.mark.integration(mi_versions=[(25, 2)])
+    @pytest.mark.xfail(reason="CR-1614", strict=True)
```
Andy-Grigg, line 267: Duplicating the tests seemed like the easiest way of xfailing only the 25.2 version of the tests. These will likely diverge anyway once we have the updated data to test against.

tests/conftest.py
Andy-Grigg: I reverted these changes and instead deleted the broken sections of the example. This is because the docs build would still fail. We can revert this deletion once the underlying API issue is resolved.

tests/test_integration_tests.py
ludovicsteinbach: You can call xfail in the test method (Second example in this section https://docs.pytest.org/en/6.2.x/skipping.html#xfail-mark-test-functions-as-expected-to-fail), and I assume there is a way to programmatically read the mark from the test as well.

But as you say it is likely to diverge anyway

examples/4_Sustainability/4-1_BoM_Sustainability_summary.py
```diff
@@ -49,320 +49,3 @@
     .with_units(mass=MASS_UNIT, energy=ENERGY_UNIT, distance=DISTANCE_UNIT)
 )
 # -
-
```
da1910, line 52: Is this to address #687?

tests/test_integration_tests.py
Andy-Grigg: We already have a copy of the 25R2 database, so I'll prepare that this afternoon and get the server upgraded. If the tests still pass, I'll implement this as a runtime check. Otherwise, I'll keep this in place.

examples/4_Sustainability/4-1_BoM_Sustainability_summary.py
Andy-Grigg: No, that's just because deserialization of the summary type is broken. It's awkward, but possible, to add a mark to the dynamically-generated parameterization, but it's a lot more effort to avoid building the notebook. The options were:

* Comment out each line after this point. Not great, and easy to miss some uncomments because of the notebook format we use.
* Remove the reference to the notebook from sphinx, which should stop it building. You can't really do comments in ReST, so it's a deletion.

I went with deleting the problematic code. It's obvious that something's gone, so we won't forget to undelete it.

tests/test_integration_tests.py
Andy-Grigg: As far as I can tell, there are no failures for 25R2. I'll recombine these test cases.

ludovicsteinbach (approved):

da1910 (approved):

PR #719, opened 2025-02-26
Add part and process-level transport information to BoM Sustainability responses

### LINKED ISSUE #720: Support 2025 R2 updates to BomSustainability request
Add support for additional `transport_stages` to parts and processes. Extend tests to cover the new cases.

Andy-Grigg:
Closes #720 

Supports 24/12 BoM-based sustainability queries:

* 24/12 BoMs can now be provided to  `BomSustainabilityQuery.with_bom()`
* Responses with `transport_stages` are correctly added to the relevant item_result objects. The new model fields are allowed to be `Unset`, since these fields are new in the 2025 R2 version of the API.
* Added tests to validate the new behavior. The example response payloads do not include these new transport fields, so check these manually. Add integration tests to validate behavior for existing API versions and the new API version.
* Update API documentation with guidance about API release versions and BoM namespaces.


.github/workflows/server_checks.yml
```diff
@@ -151,22 +151,13 @@ jobs:
...
-
       - name: "Upload coverage to Codecov"
         uses: codecov/codecov-action@v5
-        if: ${{ matrix.os  == 'ubuntu-latest' && matrix.server == 'TEST_SERVER_DEV_URL' }}
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
         with:
           files: .cov/xml
-          flags: integration
+          flags: integration-${{ matrix.os }}-${{ matrix.server }}
```
ludovicsteinbach, line 160: An alias for matrix.server might make this easier to read on codecov, at the moment it's the name of the secret.

Is coverage for each MI version a good metric? Aren't we bound to poor coverage on older versions as we start supporting new features on newer versions?

src/ansys/grantami/bomanalytics/_item_results.py
```diff
@@ -374,6 +374,12 @@ def create_part_with_sustainability(
         part_with_sustainability._add_child_parts(_raise_if_unset(result_with_sustainability.parts))
         part_with_sustainability._add_child_materials(_raise_if_unset(result_with_sustainability.materials))
         part_with_sustainability._add_child_processes(_raise_if_unset(result_with_sustainability.processes))
+
+        # transport_stages is optional, only returned by BAS 2025 R2 and later
+        if result_with_sustainability.transport_stages:
+            part_with_sustainability._add_child_transport_stages(
+                _raise_if_unset(result_with_sustainability.transport_stages)
```
ludovicsteinbach, line 381: If it is optional, should this be using `_raise_if_unset`?

src/ansys/grantami/bomanalytics/queries.py
```diff
@@ -1875,6 +1880,6 @@ class BomSustainabilitySummaryQuery(_SustainabilityMixin, _BomQueryBuilder):
 
     """
 
-    _supported_bom_formats = [_BomFormat.bom_xml2301, _BomFormat.bom_xml2412]
+    _supported_bom_formats = [_BomFormat.bom_xml2301]
```
ludovicsteinbach, line 1883: What's happening here, does it not support 24/12?

ludovicsteinbach (approved):
Looks good, some minor questions

src/ansys/grantami/bomanalytics/queries.py
Andy-Grigg: It will, but right now the deserialization is broken for responses of requests made using that XML type.

I think I originally added this change in when I moved to the v2 BAS API. That change shouldn't have gone in then though. It will be reinstated when #721 is implemented.

src/ansys/grantami/bomanalytics/_item_results.py
Andy-Grigg: Good point. Since we're explicitly checking if it's truthy first (and therefore not `Unset`), we can't have gotten to this point with an unset value.

The 'process' version of this method already doesn't include `_raise_if_unset`.

.github/workflows/server_checks.yml
Andy-Grigg: It doesn't track coverage separately they all get aggregated together for the PR. My reason for doing this is because with this PR, we'll have tests that don't run on certain platforms at all. In future, we could have actual production code as well which can't execute either, so we need to gather coverage from all runs.

As for alias, are you thinking something like:

```yaml
        server:
          - url: "TEST_SERVER_DEV_URL"
            name: "dev"
          - url: "TEST_SERVER_25R1_URL"
            name: "2025 R1"
          - url: "TEST_SERVER_24R2_URL"
            name: "2024 R2"
```

Andy-Grigg: While actually typing this out, I realized I don't think this is a good idea. The tag in coverage only appears if you want to drill down into which runs are covering which lines. If at that level you get the name of the secret instead of the actual version number, that's fine.

PR #724, opened 2025-02-28
Add transport groupings by part and by category to sustainability summary results

### LINKED ISSUE #721: Support 2025 R2 updates to BomSustainabilitySummary request
Add new properties to sustainability summary response. Extend tests to check new behavior.

Andy-Grigg:
Closes #721

Work in progress


src/ansys/grantami/bomanalytics/_query_results.py
```diff
@@ -894,13 +915,37 @@ def process(self) -> SustainabilityPhaseSummaryResult:
...
 
-        Values in percentages express the contribution of the specific transport stage, relative to contributions of all
-        transport stages.
+        Values in percentages express the contribution of the specific transport stage, relative to the total
+        contribution from transportation in the BoM.
         """
         return self._transport_details
 
+    @property
+    def transport_details_grouped_by_category(self) -> List[TransportSummaryByCategoryResult]:
```
Andy-Grigg, line 926: I don't know if it's better for this to return an empty list or `None` if not provided by the API. If it's `None` then it's more obvious that this isn't a thing the server knows about, but it's irritating that it forces you to do a `None` check before iterating.

tests/inputs/examples.py
```diff
@@ -84,6 +84,7 @@
...
...CategorySummary&quot;:[{&quot;Category&quot;:&quot;Distribution&quot;,&quot;Distance&quot;:{&quot;Value&quot;:0,&quot;Unit&quot;:&quot;km&quot;},&quot;EmbodiedEnergy&quot;:{&quot;Value&quot;:0,&quot;Unit&quot;:&quot;MJ&quot;},&quot;EmbodiedEnergyPercentage&quot;:0,&quot;ClimateChange&quot;:{&quot;Value&quot;:0,&quot;Unit&quot;:&quot;kg&quot;},&quot;ClimateChangePercentage&quot;:0},{&quot;Category&quot;:&quot;Manufacturing&quot;,&quot;Distance&quot;:{&quot;Value&quot;:200,&quot;Unit&quot;:&quot;km&quot;},&quot;EmbodiedEnergy&quot;:{&quot;Value&quot;:4.72415244615649,&quot;Unit&quot;:&quot;MJ&quot;},&quot;EmbodiedEnergyPercentage&quot;:100,&quot;ClimateChange&quot;:{&quot;Value&quot;:0.333123822320444,&quot;Unit&quot;:&quot;kg&quot;},&quot;ClimateChangePercentage&quot;:100}],&quot;PhaseSummary&quot;:{&quot;Phase&quot;:&quot;Transport&quot;,&quot;EmbodiedEnergy&quot;:{&quot;Value&quot;:4.72415244615649,&quot;Unit&quot;:&quot;MJ&quot;},&quot;EmbodiedEnergyPercentage&quot;:0.0295650122079945,&quot;ClimateChange&quot;:{&quot;Value&quot;:0.333123822320444,&quot;Unit&quot;:&quot;kg&quot;},&quot;ClimateChangePercentage&quot;:0.0154556592789601}},&quot;LogMessages&quot;:[]}""",  # noqa: E501
```
Andy-Grigg, line 87: We probably want to rethink how this works. It's useful having a blob to assert from, but it's a pain to set up.

Andy-Grigg:
@ludovicsteinbach This is mostly complete if you want to take a look. Still needs a bit of tidying though before it's ready for a full review.

src/ansys/grantami/bomanalytics/_query_results.py
```diff
@@ -891,16 +925,82 @@ def process(self) -> SustainabilityPhaseSummaryResult:
         """
         return self._process_summary
 
+    @property
+    def distribution_transport_summary(self) -> TransportSummaryByCategoryResult | None:
```
Andy-Grigg, line 929: Since there are only ever two categories in a result, it makes sense to define two properties which each return either a result or None if using an old version of the API. This is different to the summary by job, which has a variable number of results, and so returns a list. In that case it always returns an empty list, but the list may be empty.

tests/inputs/examples.py
Andy-Grigg: I will experiment in a separate branch with using a copy and paste of the yaml pulled straight from the API definition.

src/ansys/grantami/bomanalytics/_item_results.py
```diff
@@ -1893,18 +1977,77 @@ def transport_reference(self) -> TransportReference:
...
+
+    @property
+    def category(self) -> TransportCategory:
+        """
+        The transport category for this summary.
+        """
+        return self._category
+
+    @property
+    def transport_types(self) -> set[str]:
```
Andy-Grigg, line 2044: I thought this made more sense as a set than a list, since it contains the unique set of transport types defined in the aggregation. There's no ordering and the values are unique, so list.

src/ansys/grantami/bomanalytics/_item_results.py
```diff
@@ -544,6 +552,64 @@ def create_transport_summary(
...
+        -------
+        TransportSummaryByPartResult
+        """
+
+        # BAS does not support nullable enums in server responses, so NotApplicable is used to indicate that
+        # the category should be null.
+        if result.category == "NotApplicable":
+            category = None
+        else:
+            TransportCategory(result.category)  # type: ignore[call-arg]
```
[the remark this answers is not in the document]
Andy-Grigg: Codacy is also complaining about this, but I don't see any alternative apart from just not documenting the enums.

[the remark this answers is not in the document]
ludovicsteinbach: If I understand what we're trying to do here, I believe this works:
Code
```
from enum import Enum
class TransportCategory(Enum):
    """The stage of the product lifecycle to which a :class:`.TransportSummaryByPartResult` belongs."""

    MANUFACTURING = "Manufacturing"
    """Transportation of individual components before the product is completed."""

    DISTRIBUTION = "Distribution"
    """Transportation of the complete finished product."""
```
RST
```
.. autoclass:: ansys.grantami.bomanalytics._item_results.TransportCategory
   :members:
   :undoc-members:
```

Html
image

I'm not sure where the constructor args come from. We definitely managed to document enums nicely for RecordLists (https://recordlists.grantami.docs.pyansys.com/version/dev/api/models.html#ansys.grantami.recordlists.UserRole)
This could be some extension, or autodoc setting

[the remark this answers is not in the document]
Andy-Grigg: I think the constructor args can be removed by doing `.. autoclass:: ansys.grantami.bomanalytics._item_results.TransportCategory()`, with `()` after the class name. This is good though, I don't think this was possible when bomanalytics was first released.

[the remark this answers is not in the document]
ludovicsteinbach: https://peps.python.org/pep-0257/  String literals occurring immediately after a simple assignment at the top level of a module, class, or __init__ method are called “attribute docstrings”.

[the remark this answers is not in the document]
Andy-Grigg: I tried that as a 'hail mary' when this mypy error occurred, but they didn't show up. I think `.. autoenum::` doesn't handle them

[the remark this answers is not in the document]
ludovicsteinbach: > I think the constructor args can be removed by doing `.. autoclass:: ansys.grantami.bomanalytics._item_results.TransportCategory()`, with `()` after the class name. This is good though, I don't think this was possible when bomanalytics was first released.

Ah yes. But recordlists still document UserRole(value), I think something is different but I can't find what. Removing the constructor args entirely makes sense though

src/ansys/grantami/bomanalytics/_query_results.py
```diff
@@ -891,16 +925,82 @@ def process(self) -> SustainabilityPhaseSummaryResult:
...
+
+        Notes
+        -----
+        This property is only populated if the Granta MI servers includes MI Restricted Substances and Sustainability
+        Reports 2025 R2 and later. Otherwise, this property will return ``None``.
+        """
+        return self._distribution_transport_summary
+
+    @property
+    def manufacturing_transport_summary(self) -> TransportSummaryByCategoryResult | None:
```
ludovicsteinbach, line 948: Does this need the note about input BoM version? Or does 2025R2 reports return this property for old boms?

ludovicsteinbach (approved):
Looks good. I'd like it if we could remove the Enum weirdness though

src/ansys/grantami/bomanalytics/_query_results.py
Andy-Grigg: It does, it's just full of zeros. This is the same behavior if you have a 24/12 BoM with no manufacturing transport stages, e.g. by taking a 23/01 BoM and upgrading it to 24/12. I didn't think it was worth documenting this case specifically.

src/ansys/grantami/bomanalytics/_item_results.py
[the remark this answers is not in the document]
Andy-Grigg: These changes are made for existing Enums here https://github.com/ansys/grantami-bomanalytics/pull/726.

Once that PR is done, I'll merge here and make the changes suggested above.

tests/inputs/examples.py
Andy-Grigg: https://github.com/ansys/grantami-bomanalytics/pull/727

PR #727, opened 2025-03-06
Re-organize test BoMs and payloads

Andy-Grigg:
Refactor how we use example BoMs and response payloads in tests.

This started out with trying to use the yaml examples directly as copy-and-paste from the swagger documentation, to give more flexibility in terms of which samples we use in which situations. However, it then made sense to refactor the XML BoM examples as well.

I think this new format is easier to use and more flexible if we ever want to extend our suite of mocked tests. But this is highly subjective, so I'm interested in feedback.


tests/inputs/payloads/GetComplianceForBom.Response.yaml
```diff
@@ -0,0 +1,53 @@
+Parts:
```
ludovicsteinbach, line 1: How did you put these files together? Manually copied from the definition?

tests/inputs/prepare_boms.py
```diff
@@ -0,0 +1,43 @@
...
+_bom_dir = _inputs_dir / "boms"
+
+
+@dataclass(frozen=True)
+class BoM:
+    path: pathlib.Path
+    content: str
+
+
+example_boms: dict[str, BoM] = {}
```
ludovicsteinbach, line 36: In general I'd say we should avoid imports with side effects.
An alternative here would to encapsulate this logic in a method and use a fixture to expose the result. However looking at how it is used that would require perhaps a bit more restructuring work to consume the fixture in tests.

pyproject.toml
```diff
@@ -66,6 +66,7 @@ plotly = ">=5.17,<6.0"
 ipywidgets = "^8.1.1"
 mypy = "^1.9.0"
 defusedxml = "^0.7"
+ruamel_yaml = "^0.18"
```
ludovicsteinbach, line 69: Is there any particular feature offered by this library that other yaml parsing library would not provide?

tests/inputs/payloads/GetComplianceForBom.Response.yaml
Andy-Grigg: Yes, literally copied the example and de-indented. I did wonder if we wanted to embed the entire definition and access the examples, but that seemed like more code for minimal effort. These files don't need to be examples from the definition though, in future we could prepare anything we need.

pyproject.toml
Andy-Grigg: To be honest, I googled around and went with the one that I saw the most references for. For what we're doing, I don't think it matters.

Do we have a preferred library?

tests/inputs/prepare_boms.py
Andy-Grigg: This has turned into a bit of a quagmire. As you say, will probably mean significantly re-writing most of the mocked tests. I'll have another look next week, but this might be more hassle than it's worth.

pyproject.toml
Andy-Grigg: Switched out to pyyaml, which is used by ansys-sphinx-theme and jupytext already.

ludovicsteinbach (approved):

Issue #722, opened 2025-02-26
Extend example notebooks to cover 24/12 sustainability BoM functionality


Andy-Grigg:
Extend notebooks to include additional transport response properties

PR #728, opened 2025-03-07
Update examples to include transport results

Andy-Grigg:
#722 

Update examples to include new transport results:

* Extend sustainability summary example to cover new transport aggregations by part and by category, including new plots
* Extend sustainability example to cover new child transport results

In both cases, change the BoM to be a 24/12 format BoM, and include a note both at the top of the page about the change, and throughout the example if the results would be appreciably different.


examples/4_Sustainability/4-1_BoM_Sustainability_summary.py
```diff
@@ -135,10 +151,32 @@ def plot_impact(df, title, textinfo="percent+label", hoverinfo="value+name"):
...
 )
+transport_df_full.describe()
+# -
+
+# Most of these transport stages contribute little to the overall sustainability impact. To make a visualization more
+# insightful, group all transport stages that contribute less than 5% of embodied energy or climate change in a single
+# 'Other' transport stage.
+
+# +
+# Define the criterion
```
Andy-Grigg, line 162: We could consider adding this to the library. We could include a 'get_aggregated_transport_details()` method or something which accepts a threshold and does the filtering.

Andy-Grigg:
Discovered #729 while working on this

tests/conftest.py
```diff
@@ -96,12 +91,34 @@ def pytest_generate_tests(metafunc):
     'example_script' fixture.
     """
 
+    mi_version = get_mi_version()
```
Andy-Grigg, line 94: This is all getting a bit absurd. We could just pass an environment variable in, and assume the latest version if it's not available.

examples/4_Sustainability/4-2_BoM_Sustainability.py
```diff
@@ -112,20 +128,28 @@
...
-# from database data and masses defined in the BoM.
+# Processes are represented by ``ProcessWithSustainabilityResult`` objects. They include the
+# ``.transport_stages`` property. This property consists of the transportation of the part
+# during the manufacturing process, and is defined as a list of ``TransportWithSustainabilityResult``
+# objects.
+#
+# The environmental impact of a process is calculated from database data and the dimensional details of
+# the process defined in the BoM.
+#
+# Processes may appear as children of processes in the hierarchy, but the environmental impact of
```
ludovicsteinbach, line 145: Processes are nestable?

tests/conftest.py
```diff
@@ -96,12 +91,34 @@ def pytest_generate_tests(metafunc):
...
+
+    # If we don't have an MI version, we cannot contact an MI server. Example tests cannot run, so we might as well
+    # return early.
+    if mi_version is None:
+        return
+
+    xfail_examples_for_version = defaultdict(set)
+    xfail_examples_for_version[(25, 1)] = {"4-1_BoM_Sustainability_summary.py", "4-2_BoM_Sustainability.py"}
+    xfail_examples_for_version[(24, 2)] = {"4-1_BoM_Sustainability_summary.py", "4-2_BoM_Sustainability.py"}
```
ludovicsteinbach, line 103: Why do these need to be xfailed?

Andy-Grigg: Because these examples now use 24/12 BoMs by default, which don't succeed on older versions of BAS.

We could create compatible versions or manipulate the example before executing it. Both of those options seem complicated though.

examples/4_Sustainability/4-2_BoM_Sustainability.py
Andy-Grigg: Whoops, that's a copy and paste error. Thank you for catching!

tests/conftest.py
ludovicsteinbach: Ah right, I was offset in what the latest version was ! Makes sense.
I was about to suggest that we could skip them instead but it looks like that's what's being done. Could the config variable names reflect that?

ludovicsteinbach (approved):

PR #738, opened 2025-03-12
Documentation review

EstherProsperi:


doc/source/conf.py
```diff
@@ -115,7 +115,7 @@
 
 extlinks = {
     'MI_docs': (
-        'https://ansyshelp.ansys.com/account/secured?returnurl=/Views/Secured/Granta/v242/en/RS_and_Sustainability/%s',
+        'https://ansyshelp.ansys.com/account/secured?returnurl=/Views/Secured/Granta/v251/en/RS_and_Sustainability/%s',
```
EstherProsperi, line 118: we should probably update this to the 25 R2 path:

https://ansyshelp.ansys.com/account/secured?returnurl=/Views/Secured/Granta/v252/en/Granta_MI/one_mi/ (note the folder change here)

You can check the paths we need on the qa site before they go live:
https://ansysproducthelpqa.win.ansys.com/account/secured?returnurl=/Views/Secured/Granta/v252/en/Granta_MI/one_mi/bom_analyzer.html

Andy-Grigg: We'll need to make the change to v252 after release, since this package will need to be released before the 2025 R2 docs go live. I'll create a with a 2025 R2 fix version to remind us to put out a patch release once the docs are live.

This is an awkward side-effect of needing to link to docs. The alternative is that we publish with broken links for a few days, but that isn't great either.

Andy-Grigg: See https://github.com/ansys/grantami-bomanalytics/issues/739

Andy-Grigg (approved):

PR #748, opened 2025-03-25
Allow dependabot server checks to run in parallel

Andy-Grigg:
This change should allow dependabot checks to run fully in parallel:

* If the PR HEAD ref starts with 'dependabot/', then include the ref in the concurrency group, to make the group unique.
* If the PR HEAD ref does not start with 'dependabot/', then don't include the ref (existing behavior).

Dependabot doesn't actually run server checks, but they still need to run through this step so we can ensure they were successfully skipped.


Andy-Grigg:
Another alternative is to just remove this concurrency check altogether. @da1910 I welcome your thoughts here.

da1910 (approved):

PR #751, opened 2025-03-31
Update bomanalytics-openapi to 4.0.0.dev165

Andy-Grigg:
Update bomanalytics-openapi to 4.0.0.dev165

The underlying version of BAS introduces additional nullable properties. Unfortunately, since we disambiguate between null properties and omitted properties in the response, we now need to handle this case in additional places.

These changes were driven by ensuring mypy checks passed. No significant runtime changes expected.


Andy-Grigg:
The codecov error is valid - we don't have any unit tests that validate if a response with Unset and/or None values is handled properly. I'll add some cases in.

src/ansys/grantami/bomanalytics/_typing.py
```diff
@@ -38,17 +38,18 @@ def _raise_if_unset(value: Union[T, Unset_Type]) -> T:
...
-        If the object is an instance of Unset_Type.
+        If the object is an instance of Unset_Type or is None.
     """
-    new_value = _convert_unset_to_none(value)
-    if new_value is None:
+    if value is None:
         raise ValueError("Provided value cannot be 'Unset'")
-    return new_value
+    elif isinstance(value, Unset_Type):
+        raise ValueError("Provided value cannot be 'None'")
```
ludovicsteinbach, line 51: I think errors and conditions are mismatched. Maybe a little unit test to keep us honest wouldn't hurt

Andy-Grigg:
I originally created some yaml test cases to cover these Unset and None values in real responses, but I realized in doing this that we probably should make our package handle additional API responses in more cases. I created #753 to cover this.

ludovicsteinbach (approved):

PR #773, opened 2025-05-05
Generate provenance attestations

### LINKED ISSUE #767: Add attestations
Add attestations to CI. See https://github.com/ansys/grantami-dataflow-extensions/pull/115/files for an example

ludovicsteinbach:
Closes #767 

Add attestation provenance generation to wheelhouse and build jobs.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -59,7 +59,7 @@ jobs:
           documentation-html
           documentation-pdf
 
-  smoke-tests:
+  build-wheelhouse:
     name: "Build wheelhouse for latest Python versions"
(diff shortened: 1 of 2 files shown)


Andy-Grigg (approved):

Andy-Grigg:
There's something funky going on with the pre-merge style checks. It looks like the pre-commit environment for the ansys pre-commit hooks is failing to build because `wheel` doesn't exist on private PyPI.

A new version of wheel was released and yanked within the last month or so, maybe private PyPI doesn't handle this situation properly when a release is yanked?

Andy-Grigg:
According to GitHub this was yanked almost a month ago, so probably not the direct cause.

PR #787, opened 2025-05-12
Update bomanalytics-openapi dependency to 4.0.0 stable release

Andy-Grigg:
Use a semver-style dependency specification to allow for bug fixes and upgrades to bomanalytics-openapi.


ludovicsteinbach (approved):
I think that's sensible.

pyproject.toml
```diff
@@ -40,7 +40,8 @@ python = "^3.10.0"
 
 # Packages for core library
 ansys-openapi-common = "^2.0.0"
-ansys-grantami-bomanalytics-openapi = "4.0.0rc4"
+# Ensure PIP_INDEX_URL is not set in CI when building on release branches.
+ansys-grantami-bomanalytics-openapi = "^4.0.0"
```
da1910, line 44: Two things occur to me - firstly poetry is moving away from caret dependencies, and secondly do we want to support a hypothetical 4.1? If so we would need "^4.0" if my reading of https://python-poetry.org/docs/dependency-specification/ is correct

Andy-Grigg: I didn't realize, in which case we probably should update to use `~=` whenever we make a change.

As for the current behavior, I look at the first row of https://python-poetry.org/docs/dependency-specification/#caret-requirements. Assuming 0 minor and patch versions have no special meaning (which I don't think they do), I don't think it makes a difference.

It does make a difference for `~=` though. Here I believe we want `~=4.0`.

Andy-Grigg: I have updated to use an equivalent compatible requirement.

See https://github.com/ansys/grantami-bomanalytics/pull/787#discussion_r2084986943 for a longer response.

da1910 (approved):

PR #792, opened 2025-05-14
Shutdown 2025R2 machine

Andy-Grigg:
Whilst main currently does not start the 2025 R2 release machine, it may be running when main CI starts. We need to try and shut it down in case it is running.

### DIFF
--- a/.github/workflows/server_checks.yml
+++ b/.github/workflows/server_checks.yml
@@ -28,6 +28,8 @@ on:
       # Test server names
       AZURE_VM_NAME_DEV:
         required: true
+      AZURE_VM_NAME_25R2:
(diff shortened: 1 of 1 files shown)


ludovicsteinbach (approved):
Is the idea to use the dev machine as the 25R2 machine for main until it can be updated to 26R1?

Andy-Grigg:
> Is the idea to use the dev machine as the 25R2 machine for main until it can be updated to 26R1?

I started responding, and then realized the point of what you were saying. This PR needs to add tests on 25R2 as well explicitly. I'll update.

ludovicsteinbach (approved):

PR #803, opened 2025-05-29
Sankey diagram example

### LINKED ISSUE #736: Example proposal: Sankey diagram
### 📝 Description of the example

We already have an example that shows the hierarchy in graphical form as a hierarchical pie chart, or sunburst chart. We should be able to instead use this data to plot a Sankey diagram, which is a common way of visualizing sustainability data.

### 📁 Files needed for running the example

https://plotly.com/python/sankey-diagram/

### 💻 Which operating system are you using?

Windows

### 📀 Which ANSYS version are you using?

2025 R2

### 🐍 Which Python version are you using?

3.12

### 📦 Installed packages

```shell
N/A
```

Andy-Grigg:
Closes #736 

Add an example of a Sankey diagram. This is still somewhat limited, there is no in-use, circularity, or end-of-life aspect to the analytics, but nevertheless this shows the principle of how the total embodied energy of a product is created based on the materials, processes, and transportation.

This could also be done as part of the full hierarchical BoM. I might have a look to see how the visualization works there, but I feel like the data would be too complex to make sense of.

I also thought about adding this as a standalone example. I don't think it warrants this on its own, but possibly the graphing could all be broken out into individual examples.

### DIFF
--- a/examples/4_Sustainability/4-1_BoM_Sustainability_summary.py
+++ b/examples/4_Sustainability/4-1_BoM_Sustainability_summary.py
@@ -424,10 +424,12 @@ plot_impact(
     textinfo="percent", hoverinfo="value+name+label"
 )
 
-# ## Hierarchical view
+# ## Data visualization
 #
-# Finally, aggregate the sustainability summary results into a single dataframe and present it in a hierarchical
-# chart. This highlights the largest contributors at each level. In this example, two levels are defined:
+# ### Tabulated data
+#
+# To aid with data visualization, aggregate the sustainability summary results into a single dataframe and present it in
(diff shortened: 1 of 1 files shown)


da1910:
Looks fine in terms of the code and it renders the code cells correctly - nothing is showing up for the plots for me locally, just a warning that ploty objects cannot be displayed. I suspect this is not a problem with this PR

da1910 (approved):

Andy-Grigg:
I have spiked splitting the Sustainability Summary example into separate notebooks for the summary, materials and processes, transports, and full hierarchy on a separate branch. I'll try and finish that off today, but it will be a separate branch and PR.

PR #819, opened 2025-06-05
Add an example of creating a BoM from a CSV file

### LINKED ISSUE #733: Example request: Create BoM from CSV
We have an example creating a BoM from a JSON format, but not from a CSV. This could go via a dataframe, showing that once a BoM-type structure is in a dataframe with certain specific columns, that those can be used to create a hierarchy of `bom_types` objects which can be converted into an XML BoM.

We could even consider adding this functionality to the library itself as a BoM handler extension. If the user can provide a dataframe in a specific format, we could convert that into a BoM. This could be considered the 'reverse' of #731.

ludovicsteinbach:
Close #733 

Add example of processing a CSV file into a BoM.
CSV format matches the output of BoMAnalyzer reports: sequential traversal of the bom, no direct reference to the parent, but a value that represents the level in the BoM.

### DIFF
--- a/examples/6_BoMs/6-3_Creating_an_XML_BoM_from_CSV.py
+++ b/examples/6_BoMs/6-3_Creating_an_XML_BoM_from_CSV.py
@@ -15,8 +15,8 @@
 
 # # Create an XML BoM from a CSV data source
 
-# This example shows how to use the ``bom_types`` subpackage to create a valid Granta MI XML
-# BoM. This subpackage can be used to help construct a Granta 24/12-compliant XML BoM file to
+# This example shows how to use the ``bom_types`` subpackage to create a Granta MI BoM.
+# This subpackage can be used to help construct a Granta 24/12-compliant XML BoM file to
 # use with the BoM queries provided by this package. The code in this example shows how to generate
 # a BoM from a representative CSV data source. The general approach can be applied to data
 # in other formats or provided by other APIs.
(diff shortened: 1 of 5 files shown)


ludovicsteinbach:
I have no idea why it decided to change the changelog entry from documentation to miscellanous. I'm guessing it has to do with the update to v10 of the actions?

Andy-Grigg:
> I have no idea why it decided to change the changelog entry from documentation to miscellanous. I'm guessing it has to do with the update to v10 of the actions?

That is strange. We can fix it with a direct push to main after merge I suppose

Andy-Grigg (approved):
I think these changes look good. I've requested a review from Esther though before merging.

examples/6_BoMs/6-3_Creating_an_XML_BoM_from_CSV.py
```diff
@@ -0,0 +1,180 @@
...
+# The CSV file describes the bill of materials for the door assembly introduced in the
+# [Creating an XML BoM](./6-1_Creating_an_XML_BoM.ipynb) example.
+# The door (24X6-30) contains two hinges (HA-42-Al) and a handle (H-S-BR-Dual), both fixed to the panel (P-30-L) with
+# machine screws (DIN-7991-M8-20) and washers (N0403.12N.2). The door glass (321-51) is coated with a partially
+# reflective polymer film (7000001298).
+#
+# The hierarchy of items within the BoM is defined by the order of items in the CSV and the ``BoM Level`` field. If an
+# item has the BoM level *n*, then the item's parent is the first level preceding it in the CSV with a BoM level *n-1*.
+#
+# Each item includes an ``Item Type`` field that identifies the type of the item, here only ``Part`` or ``Material``.
```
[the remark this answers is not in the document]
EstherProsperi: Generally the use of fields here is making me twitchy - not sure if it should (but we're quite intentional about not calling them that in the help). Am making this comment here and assuming the outcome will need to be reflected throughout the topic

examples/6_BoMs/6-3_Creating_an_XML_BoM_from_CSV.py
```diff
@@ -0,0 +1,180 @@
...
+# Additional fields are specific to the type of item.
+#
+# ### Part items
+#
+# Items that refer to parts only exist in the BoM and do not reference records in Granta MI. Their ``Name`` and ``ID``
+# are defined only in the BoM to identify parts.
+#
+# The ``Quantity`` and ``Unit of measure`` fields describe the quantity of part expected in the parent. This can be
```
EstherProsperi, line 59:
```suggestion
# ``Quantity`` and ``Unit of measure`` describe the quantity of part expected in the parent. This can be
```

examples/6_BoMs/6-3_Creating_an_XML_BoM_from_CSV.py
```diff
@@ -0,0 +1,180 @@
...
+# done by specifying how many occurrences of a part are in an assembly, or for example for the glass door, the surface
+# area of laminated glass.
+#
+# There are three types of components in the BoM:
```
EstherProsperi, line 63: in _this_ BoM? looks like we have more possibilities: 

https://miniature-fortnight-dc09be55.pages.github.io/granta/v261/en/Granta_MI/one_mi/bom_items_rs.html

https://miniature-fortnight-dc09be55.pages.github.io/granta/v261/en/Granta_MI/one_mi/bom_types_sustainability.html

examples/6_BoMs/6-3_Creating_an_XML_BoM_from_CSV.py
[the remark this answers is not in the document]
ludovicsteinbach: This was more about the content of the CSV file, than the content of an MI BoM. I realise that "field" isn't really commonly used for CSVs. If I were to changes that to column/columns would that work or do you still recommend just removing it?

examples/6_BoMs/6-3_Creating_an_XML_BoM_from_CSV.py
ludovicsteinbach: Definitely only talking about this particular BoM

EstherProsperi (approved):

PR #865, opened 2025-08-15
Upgrade plotly, add default renderer

### LINKED ISSUE #807: Notebook to HTML conversion not rendering plots with plotly >=6.1
Split from #804 

Plotly >=6.1 seems to be incompatible with how we render charts in HTML examples. A temporary fix is to downgrade plotly to 6.0.x, but it would be good to understand what's going on here.

Andy-Grigg:
Closes #807 

Upgrade plotly and override the default renderer in the ipython config to `"notebook+plotly_mimetype+pdf"`. This generates compatible outputs for:

* Interactive jupyter lab usage, via the `plotly_mimetype` renderer
* HTML outputs as part of the nbconvert doc build process, using the `notebook` renderer, which embeds the plotly js in the HTML output
* PDF outputs, which generates static PDF versions of the plots using kaleido and chromium, and embeds the results in the PDF

To make the PDF output work in a portable way, during ipython startup we attempt to initialize a kaleido object. If this fails due to missing chrome, we install chrome using the required function.

In the linked issue, I expressed concern that this fix seemed non-ideal. However, the plotly docs currently state that this is the recommended approach for our purposes. See https://plotly.com/python/renderers/#multiple-renderers and https://plotly.com/python/renderers/#interactive-renderers for the relevant passages.


pyproject.toml
```diff
@@ -68,6 +68,8 @@ ipywidgets = "^8.1.1"
 mypy = "^1.9.0"
 pyyaml = "^6"
 types-defusedxml = "^0.7"
+kaleido = "^1.0"
```
ludovicsteinbach, line 71: Should these be in the doc group?

PR #873, opened 2025-08-26
Re-work allowed_types module to support multiple arguments

Andy-Grigg:
This PR re-works the `validate_argument_type` decorator to accept an initial required argument of `argument_name`, which defines which argument will be validated by the decorator instance.

This is required by a separate in-progress PR which will introduce additional arguments to type-checked methods. In these cases, the decorator will be called multiple times for each argument that is to be type checked.

As part of this work, some checks which do not depend on the actual values are now performed when the decorator is initialized. Specifically, checking that unordered collections contain a single item are now done up-front, and raise ValueErrors in these cases.

I did look at using typeguard, but this suffers from the same problem that it did when bomanalytics was first written, in that `str` is a valid type for an `Iterable[str]` type hint. As a result, I have stuck with the custom implementation.

### DIFF
--- a/src/ansys/grantami/bomanalytics/_allowed_types.py
+++ b/src/ansys/grantami/bomanalytics/_allowed_types.py
@@ -31,40 +31,51 @@ T
 """
 
 import functools
+import inspect
 from typing import Any, Callable, Type, TypeVar
 
 T = TypeVar("T")
 
 
-def validate_argument_type(*allowed_types: Any) -> Callable:
-    """Validates the types of values passed into a method. This decorator function accepts one or more
-    objects against which types are checked.
+def validate_argument_type(argument_name: str, *allowed_types: Any) -> Callable:
+    """
+    Validates the types of values passed into a callable.
+
+    This decorator function accepts an argument name and one or more objects against which types are checked.
 
     * If one object is provided, the value must be of that object's type.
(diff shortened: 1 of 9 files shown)


src/ansys/grantami/bomanalytics/_allowed_types.py
```diff
@@ -75,21 +86,44 @@ def validate_argument_type(*allowed_types: Any) -> Callable:
...
-    def decorator(method: Type[T]) -> Callable[[Any, Any], T]:
-        @functools.wraps(method)
+    for allowed_type in allowed_types:
+        if isinstance(allowed_type, (dict, set)):
+            if len(allowed_type) != 1:
+                raise ValueError(
+                    f"{type(allowed_type)} must contain exactly 1 item. '{allowed_type}' has length {len(allowed_type)}"
+                )
+
+    def decorator(callable_: Type[T]) -> Callable[[Any, Any], T]:
```
ludovicsteinbach, line 96: It has not been added in this PR but I'm struggling to understand type hints here. This decorator has no effect on the method signature or behavior, so I would expect the signature to be:
```
def decorator(callable_: Callable[[Any, Any], T]) -> Callable[[Any, Any], T]:
```

ludovicsteinbach (approved):

ludovicsteinbach (approved):

PR #875, opened 2025-09-04
Update test database creation scripts

Andy-Grigg:
Includes a new script to create a 'foreign RS database' to be used to validate new cross-database link capabilities added as part of #874. Also includes some refactoring of the existing scripts, including extracting configuration items to their own documented module.

The new script was used to create the database added to the 2026 R1 dev machine, required for #874


cicd/2_prepare_rs_db.py
```diff
@@ -323,27 +299,24 @@ def process_database(
...
     logger.info("Loading saved RS information...")
+    INPUT_FILE_NAME = Path(__file__).parent / DATA_FILENAME
     if not INPUT_FILE_NAME.exists():
         logger.error("No saved data available. Run 'get_cleaned_db_entries.py' against an existing test DB first")
         exit(1)
     with open(INPUT_FILE_NAME, "r", encoding="utf8") as fp:
         input_data = json.load(fp)
 
-    table_browser = TableBrowser(api_client, logger)
+    table_browser = DatabaseBrowser(api_client, logger)
```
da1910, line 313: Minor thing but the naming here seems inconsistent now, might be worth renaming the object too

da1910 (approved):
The changes look reasonable enough, the increasing complexity makes me a little nervous but the service is getting more complex, I'm not sure there's much we can do about it

PR #1062, opened 2026-06-17
Convert class-scoped fixtures to class methods

Andy-Grigg:
Pytest 9.1 has deprecated the use of class-scoped fixtures defined as instance methods without `@classmethod`.

This repo hasn't been updated to use pytest 9.1 yet, but it will get updated soon. This change should mean that CI continues to pass once the update is applied.

### DIFF
--- a/doc/changelog.d/1062.fixed.md
+++ b/doc/changelog.d/1062.fixed.md
@@ -1,0 +1,1 @@
+Convert class-scoped fixtures to class methods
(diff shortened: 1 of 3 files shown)


da1910 (approved):
I dislike this

Andy-Grigg:
> I dislike this

I'm going to merge this anyway, but what specifically is your concern? The use of method fixtures, or the fact that they now explicitly have to be declared as class methods? From what I can tell from the pytest changelog, removing their use does remove some ambiguity.

PR #1093, opened 2026-07-30
Update scripts and tests for 2027 R1

Andy-Grigg:
Update scripts and tests for 2027 R1:

* Update cicd database prep scripts to use new versions of Scripting Toolkit and Server API bindings
* Update test expectations for 2027 R1 database

### DIFF
--- a/tests/examples_expectations.py
+++ b/tests/examples_expectations.py
@@ -233,7 +233,7 @@ def example_5_1_summary_and_messages() -> None:
     assert sustainability_summary.messages == []  # noqa
     assert phases_df["Name"].to_list() == ["Material", "Processes", "Transport"]  # noqa
     assert [round(i) for i in phases_df["EE%"].to_list()] == [56, 27, 17]  # noqa
-    assert [round(i) for i in phases_df["CC%"].to_list()] == [67, 19, 15]  # noqa
+    assert [round(i) for i in phases_df["CC%"].to_list()] == [66, 19, 15]  # noqa
     # Expected cells with outputs
     assert set(Out.keys()) == {3, 4, 5, 6}, str(Out)  # noqa
 
--- a/tests/test_integration_tests.py
+++ b/tests/test_integration_tests.py
@@ -1000,6 +1000,216 @@ class TestSustainabilityBomQueries2412(_TestSustainabilityBomQueries):
         assert grouped_transport_other.parent_part_name is None
         assert grouped_transport_other.category is None
 
+    @pytest.mark.integration(mi_versions=[(27, 1)])
+    def test_sustainability_summary_query_27_1(self, connection):
+        query = queries.BomSustainabilitySummaryQuery()
+        query.with_bom(self.bom)
+        response = connection.run(query)
+
+        assert not response.messages, "\n".join([f"{m.severity}: {m.message}" for m in response.messages])
+
+        assert response.process.name == "Processes"
+        assert response.material.name == "Material"
+        assert response.transport.name == "Transport"
+        assert len(response.phases_summary) == 3
+
+        # Check overall percentages add up
+        self._check_percentages_add_up(response.phases_summary)
+
+        # Check expected summaries for materials
+        assert len(response.material_details) == 3
+        material_names = [m.identity for m in response.material_details]
+        expected_material_names = ["beryllium-beralcast191-cast", "stainless-astm-cn-7ms-cast", "steel-1010-annealed"]
+        assert all(expected_name in material_names for expected_name in expected_material_names)
+        self._check_percentages_add_up(response.material_details)
+
+        # Spot check one material summary
+        beryllium_summary = next(m for m in response.material_details if m.identity == "beryllium-beralcast191-cast")
+        assert len(beryllium_summary.contributors) == 1
+        assert beryllium_summary.contributors[0].name == "Component 1D"
+        assert beryllium_summary.contributors[0].part_number == "Part1.D"
+        assert beryllium_summary.contributors[0].material_mass_before_processing.value == pytest.approx(0.027)
+        assert beryllium_summary.mass_after_processing.value == pytest.approx(0.024)
+        assert beryllium_summary.mass_before_processing.value == pytest.approx(0.027)
+        assert beryllium_summary.material_reference.record_guid is not None
+        assert beryllium_summary.climate_change.value == pytest.approx(15.52, DEFAULT_TOLERANCE)
+        assert beryllium_summary.climate_change_percentage == pytest.approx(49.13, DEFAULT_TOLERANCE)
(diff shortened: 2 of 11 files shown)


tests/test_integration_tests.py
```diff
@@ -1000,6 +1000,216 @@ def test_sustainability_summary_query_25_2_26_1(self, connection):
         assert grouped_transport_other.parent_part_name is None
         assert grouped_transport_other.category is None
 
+    @pytest.mark.integration(mi_versions=[(27, 1)])
```
Andy-Grigg, line 1003: These are just copies of the 26R2 tests, except for specific differences called out below

tests/test_integration_tests.py
```diff
@@ -1000,6 +1000,216 @@ def test_sustainability_summary_query_25_2_26_1(self, connection):
...
+        beryllium_summary = next(m for m in response.material_details if m.identity == "beryllium-beralcast191-cast")
+        assert len(beryllium_summary.contributors) == 1
+        assert beryllium_summary.contributors[0].name == "Component 1D"
+        assert beryllium_summary.contributors[0].part_number == "Part1.D"
+        assert beryllium_summary.contributors[0].material_mass_before_processing.value == pytest.approx(0.027)
+        assert beryllium_summary.mass_after_processing.value == pytest.approx(0.024)
+        assert beryllium_summary.mass_before_processing.value == pytest.approx(0.027)
+        assert beryllium_summary.material_reference.record_guid is not None
+        assert beryllium_summary.climate_change.value == pytest.approx(15.52, DEFAULT_TOLERANCE)
+        assert beryllium_summary.climate_change_percentage == pytest.approx(49.13, DEFAULT_TOLERANCE)
```
Andy-Grigg, line 1036: Used to be 48.52

tests/test_integration_tests.py
```diff
@@ -1392,6 +1838,98 @@ def test_sustainability_query_25_1_to_26_1(self, connection):
...
+        # Subassembly
+        subassembly = product.parts[0]
+        assert len(subassembly.parts) == 2
+        assert len(subassembly.processes) == 1
+        assert not subassembly.materials
+
+        assert subassembly.input_part_number == "Part1.1[SubAssembly]"
+        assert subassembly._reference_value is None
+        assert subassembly.reported_mass.value == pytest.approx(1.45, DEFAULT_TOLERANCE)
+        assert subassembly.climate_change.value == pytest.approx(17.53, DEFAULT_TOLERANCE)
```
Andy-Grigg, line 1872: Used to be 17.93

tests/test_integration_tests.py
```diff
@@ -1300,6 +1635,117 @@ def test_sustainability_summary_query_25_1_to_26_1(self, connection):
...
+        beryllium_summary = next(m for m in response.material_details if m.identity == "beryllium-beralcast191-cast")
+        assert len(beryllium_summary.contributors) == 1
+        assert beryllium_summary.contributors[0].name == "Part1.D"
+        assert beryllium_summary.contributors[0].part_number == "Part1.D[LeafPart]"
+        assert beryllium_summary.contributors[0].material_mass_before_processing.value == pytest.approx(0.027)
+        assert beryllium_summary.mass_after_processing.value == pytest.approx(0.024)
+        assert beryllium_summary.mass_before_processing.value == pytest.approx(0.027)
+        assert beryllium_summary.material_reference.record_guid is not None
+        assert beryllium_summary.climate_change.value == pytest.approx(15.52, DEFAULT_TOLERANCE)
+        assert beryllium_summary.climate_change_percentage == pytest.approx(49.13, DEFAULT_TOLERANCE)
```
Andy-Grigg, line 1671: Used to be 48.52

tests/test_integration_tests.py
```diff
@@ -1033,7 +1243,132 @@ def test_sustainability_query_25_2_26_1(self, connection):
...
+        subassembly = product.parts[0]
+        assert len(subassembly.parts) == 2
+        assert len(subassembly.processes) == 1
+        assert not subassembly.materials
+
+        assert subassembly.input_part_number == "Part1.1"
+        assert subassembly.name == "Subassembly"
+        assert subassembly._reference_value is None
+        assert subassembly.reported_mass.value == pytest.approx(1.45, DEFAULT_TOLERANCE)
+        assert subassembly.climate_change.value == pytest.approx(20.40, DEFAULT_TOLERANCE)
```
Andy-Grigg, line 1371: Used to be 20.79

tests/examples_expectations.py
```diff
@@ -233,7 +233,7 @@ def example_5_1_summary_and_messages() -> None:
     assert sustainability_summary.messages == []  # noqa
     assert phases_df["Name"].to_list() == ["Material", "Processes", "Transport"]  # noqa
     assert [round(i) for i in phases_df["EE%"].to_list()] == [56, 27, 17]  # noqa
-    assert [round(i) for i in phases_df["CC%"].to_list()] == [67, 19, 15]  # noqa
+    assert [round(i) for i in phases_df["CC%"].to_list()] == [66, 19, 15]  # noqa
```
ludovicsteinbach, line 236: We know skip this example for previous MI versions because the expectation changed for 27R1. Is that correct?
It's been a while since I've looked at anything pygranta: Do we never have slightly different expectations for different versions? It seems drastic to drop the entire example

tests/test_integration_tests.py
ludovicsteinbach: Is there no way to achieve this via parametrization instead of duplicating tests?

tests/examples_expectations.py
Andy-Grigg: The current strategy is to just drop the test for old versions if it deviates at all. We could add something more involved, but that would be an extension to what we have now.

Happy to add it as an issue to improve, but I don't want to do too much overhaul of the process just to get things working.

tests/test_integration_tests.py
Andy-Grigg: In this specific case, probably yes. It's not universal though, because in general tests may differ by more than just a single value.

Again, I went with sticking with the same approach that we initiated in previous releases. If the changes in the expectations become less significant though, it might be worth revisiting that approach.

ludovicsteinbach: Having had a look at existing tests I suppose we have done this in the past. Although most differentiation between versions is generally bigger (feature supported and full expectations asserted vs feature unsupported and assert error raised).
There is precedent for full bom analysis test differentiation in `TestSustainabilityBomQueries2301` with `test_sustainability_summary_query_25_1_to_26_1` and `test_sustainability_summary_query_24_2` but they have more differences which makes parametrization less appealing

ludovicsteinbach (approved):

tests/examples_expectations.py
Andy-Grigg: Created https://github.com/ansys/grantami-bomanalytics/issues/1095

